PR #39, opened 2025-10-22
Alias uniqueness enforced across control layer

hachall:


dwctl/migrations/019_enforce_alias_uniqueness.sql
```diff
@@ -0,0 +1,77 @@
+-- Enforce uniqueness on deployed_models.alias with backwards compatibility
+-- This migration handles existing duplicates by appending suffixes
+
+-- Step 1: Create a function to generate unique aliases for duplicates
+CREATE OR REPLACE FUNCTION generate_unique_alias(base_alias TEXT, existing_id UUID DEFAULT NULL)
```
hachall: NB - this could cause some weird effects for anyone with the control layer currently deployed and with alias conflicts, as they will need to change the model name they are hitting in their api calls. That said, they will be experiencing unpredictable effects anyway as the onwards will not be able to differentiate between the deployments (i believe all traffic will go to the first instance with the offending alias name). Any user with known alias clashes should be notified that this Alias is going to change

PR #43, opened 2025-10-24
Conversation History

pjb157:


fergusfinn:
image
These buttons overlap and the styling looks a bit off (does it use the 'Button' component?)

image
The tick icon is too close to the input box

image
If its a generative model, can we use it to summarise the chats so you get the openAI style names here. Or is that a bit of an excessive use of tokens

fergusfinn:
closing, since has fallen out of mind. We can reopen if we have headspace to bring it up to date

PR #46, opened 2025-10-29
Backend Credit API

pjb157:
Proposed API for credits interaction:

New Endpoints:

* `POST /transactions`: Create credit transaction (grant/deduct)
* `GET /transactions/{credit_id}`: Get specific transaction, if you are not billing manager then you get an extra filter on the DB to filter by your Id. So if you search for an ID that exists but is not yours you get a 404.
* `GET /transactions`: Get all transactions, with user_id={id} query parameter. If you don't have BillingManager then you can't look at anyone but yours. 

Edit:
* `GET /users`: Can add query parameter include=billing, which can return the balance for each user in the return model
* `GET /users/{user_id}`: Get user's current balance if passed include=billing as query parameter.

### DIFF
--- a/dwctl/.sqlx/query-0013944d2aa7d7080d3bc39d603d0d52a67490da9217ce9954a5fe6cd9f15a58.json
+++ b/dwctl/.sqlx/query-0013944d2aa7d7080d3bc39d603d0d52a67490da9217ce9954a5fe6cd9f15a58.json
@@ -1,6 +1,6 @@
 {
   "db_name": "PostgreSQL",
-  "query": "\n            SELECT id, user_id, transaction_type as \"transaction_type: CreditTransactionType\", amount, balance_after, previous_transaction_id, description, created_at\n            FROM credits_transactions\n            ORDER BY created_at DESC, id DESC\n            OFFSET $1\n            LIMIT $2\n            ",
+  "query": "\n            SELECT id, user_id, transaction_type as \"transaction_type: CreditTransactionType\", amount, balance_after, source_id, previous_transaction_id, description, created_at\n            FROM credits_transactions\n            ORDER BY created_at DESC, id DESC\n            OFFSET $1\n            LIMIT $2\n            ",
   "describe": {
     "columns": [
       {
@@ -30,16 +30,21 @@
       },
       {
         "ordinal": 5,
+        "name": "source_id",
+        "type_info": "Text"
+      },
+      {
+        "ordinal": 6,
         "name": "previous_transaction_id",
         "type_info": "Uuid"
       },
       {
-        "ordinal": 6,
+        "ordinal": 7,
         "name": "description",
         "type_info": "Text"
       },
       {
-        "ordinal": 7,
+        "ordinal": 8,
         "name": "created_at",
         "type_info": "Timestamptz"
       }
@@ -56,10 +61,11 @@
       false,
       false,
       false,
+      false,
       true,
       true,
       false
     ]
   },
-  "hash": "e960f57560adc59f7694a5456b943249271431b758539d25edc5d3fd85363ce1"
+  "hash": "0013944d2aa7d7080d3bc39d603d0d52a67490da9217ce9954a5fe6cd9f15a58"
 }
--- a/dwctl/.sqlx/query-0238c7dfaf7e1859710f5bd943c182b017d3dc79c6fbae15a8e56d7bb0e73615.json
+++ b/dwctl/.sqlx/query-0238c7dfaf7e1859710f5bd943c182b017d3dc79c6fbae15a8e56d7bb0e73615.json
@@ -1,6 +1,6 @@
 {
   "db_name": "PostgreSQL",
-  "query": "\n            SELECT id, user_id, transaction_type as \"transaction_type: CreditTransactionType\",\n                amount, balance_after, previous_transaction_id, description, created_at\n            FROM credits_transactions\n            WHERE id = $1\n            ",
+  "query": "\n            SELECT id, user_id, transaction_type as \"transaction_type: CreditTransactionType\", amount, balance_after, source_id, previous_transaction_id, description, created_at\n            FROM credits_transactions\n            WHERE user_id = $1\n            ORDER BY created_at DESC, id DESC\n            OFFSET $2\n            LIMIT $3\n            ",
   "describe": {
     "columns": [
       {
@@ -30,26 +30,34 @@
       },
       {
         "ordinal": 5,
+        "name": "source_id",
+        "type_info": "Text"
+      },
+      {
+        "ordinal": 6,
         "name": "previous_transaction_id",
         "type_info": "Uuid"
       },
       {
-        "ordinal": 6,
+        "ordinal": 7,
         "name": "description",
         "type_info": "Text"
       },
       {
-        "ordinal": 7,
+        "ordinal": 8,
         "name": "created_at",
         "type_info": "Timestamptz"
       }
     ],
     "parameters": {
       "Left": [
-        "Uuid"
+        "Uuid",
+        "Int8",
+        "Int8"
       ]
     },
     "nullable": [
+      false,
       false,
       false,
       false,
@@ -60,5 +68,5 @@
       false
     ]
   },
-  "hash": "e4eaf74bc28ca2bb02278b34f2d989db4a77c1bce1b44018d1ecc4a72ac96f2a"
+  "hash": "0238c7dfaf7e1859710f5bd943c182b017d3dc79c6fbae15a8e56d7bb0e73615"
 }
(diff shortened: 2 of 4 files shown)


fergusfinn:
```
POST /users/{user_id}/credits - Add/remove credits (admin_grant/admin_removal)
GET /users/{user_id}/credits/balance - Get any user's balance
GET /users/{user_id}/credits/transactions - List any user's transactions
```

I think i'd prefer an API more like:
```
POST   /transactions                    - Create credit transaction (grant/deduct)
GET    /transactions/{credit_id}        - Get specific transaction
GET    /transactions       - Get all transactions
GET    /transactions?user_id={id}       - List transactions for a user

# User balance (computed resource)
GET    /users/{user_id}/balance    - Get user's current balance
```

I think 'transactions' are the top level resource, and '/transactions' is the (name up for debate) collection of those resources. POSTing to the collection should create an object in that collection, and GET should be used to get (a query parameter filtered list of) objects in the collection. 

`/users/../balance` is then a singular resource owned by a specific user (The alternative here is to do another top level resource called 'balances', or add a 'controller' on the transactions resource - i.e. /transactions/compute_balance?user=user_id.

The value of doing it this way is that its much easier (over time) to do credits queries that don't require you to scope to a specific user. For example, if you want to later support getting all the transactions in the last 5m, you can do it with `/transactions?since=5m`, because you haven't nested transactions under users prematurely.

dwctl/migrations/021_add_credit_system.sql
```diff
@@ -0,0 +1,55 @@
...
+
+-- Credit transactions table - simple ledger of all credit movements
+CREATE TABLE credit_transactions (
+    id BIGSERIAL PRIMARY KEY,
+    user_id UUID NOT NULL REFERENCES users(id) ON DELETE CASCADE,
+
+    -- Transaction details
+    transaction_type credit_transaction_type NOT NULL,
+    amount DECIMAL(12, 8) NOT NULL,  -- Absolute value of transaction
+    balance_after DECIMAL(12, 8) NOT NULL,  -- Running balance after this transaction
```
JoshC8C7: Did we not decide against doing running balance?

dwctl/src/db/models/credits.rs
```diff
@@ -0,0 +1,45 @@
...
+    pub balance_after: Decimal,
+    pub description: Option<String>,
+}
+
+/// Database response for a credit transaction
+#[derive(Debug, Clone)]
+pub struct CreditTransactionDBResponse {
+    pub id: i64,
+    pub user_id: UserId,
+    pub transaction_type: CreditTransactionType,
```
JoshC8C7: I reckon we want to add a hook already for an external id - whether for stripe or some other payment system that people will presumably hook in (e.g. their own chargeback coding)

dwctl/src/main.rs
```diff
@@ -566,6 +566,27 @@ pub async fn build_router(state: &mut AppState, onwards_router: Router) -> anyho
...
+            "/users/current/credits/transactions",
+            get(api::handlers::credits::list_current_user_transactions),
+        )
+        // Credit management - admin endpoints (BillingManager only)
+        .route("/users/{user_id}/credits", post(api::handlers::credits::add_user_credits))
+        .route(
+            "/users/{user_id}/credits/balance",
+            get(api::handlers::credits::get_user_balance),
+        )
+        .route(
```
JoshC8C7: do we need this to be accessible?

dwctl/src/db/models/credits.rs
fergusfinn: no reason not to do this in a follow up right?

dwctl/migrations/021_add_credit_system.sql
```diff
@@ -0,0 +1,55 @@
...
+CREATE TYPE credit_transaction_type AS ENUM (
+    'purchase',          -- User purchased credits (positive)
+    'admin_grant',       -- Admin granted credits (positive)
+    'admin_removal',     -- Admin removal of credits (negative)
+    'usage'              -- Credits used for API request (negative)
+);
+
+-- Credit transactions table - simple ledger of all credit movements
+CREATE TABLE credit_transactions (
+    id BIGSERIAL PRIMARY KEY,
```
fergusfinn: We've been using UUID keys pretty much everywhere, so for consistency we should do the same here

dwctl/migrations/021_add_credit_system.sql
```diff
@@ -0,0 +1,55 @@
...
+CREATE INDEX idx_credit_transactions_created_at ON credit_transactions (created_at DESC);
+CREATE INDEX idx_credit_transactions_type ON credit_transactions (transaction_type);
+
+-- Comments
+COMMENT ON TABLE credit_transactions IS 'Simple ledger of all credit transactions';
+COMMENT ON COLUMN credit_transactions.transaction_type IS 'Type of transaction - purchase, usage or admin adjustment';
+COMMENT ON COLUMN credit_transactions.amount IS 'Absolute value of transaction amount';
+COMMENT ON COLUMN credit_transactions.balance_after IS 'Balance after this transaction';
+
+-- Make the table append-only (prevent updates and deletes)
```
fergusfinn: Is it useful to enforce this in the DB layer? Open to both options here, just curious about the justification

dwctl/migrations/022_add_billing_manager_role.sql
```diff
@@ -0,0 +1,3 @@
+-- Add BillingManager role for credit system administration
+
+ALTER TYPE user_role ADD VALUE 'BILLINGMANAGER';
```
fergusfinn: although - if this migration just works, i might be wrong about enums. I've had some sort of problem with them before, but if we can just simply migrate them then that's fine

dwctl/src/api/models/transactions.rs
```diff
@@ -0,0 +1,77 @@
...
+use utoipa::{IntoParams, ToSchema};
+
+// Request models
+#[derive(Debug, Clone, Serialize, Deserialize, ToSchema)]
+pub struct CreditTransactionCreate {
+    /// User ID (required - UUID format)
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[schema(value_type = Option<String>, format = "uuid")]
+    pub user_id: Option<UserId>,
+    /// Type of transaction (only admin_grant and admin_removal allowed for admin API)
```
fergusfinn: If this is the case, and this is useful, then should we use a different enum here with just the correct variants and an Into<CreditTransactionType> implementation (really a From...)

dwctl/migrations/021_add_credit_system.sql
pjb157: Yep will change

dwctl/src/api/models/transactions.rs
pjb157: Okay yeah more helpful from the API schema as well. Was just in case we wanted to change in future but updating the enums in rust is cheap

dwctl/src/api/handlers/transactions.rs
```diff
@@ -0,0 +1,785 @@
...
+        ("X-Doubleword-User" = [])
+    )
+)]
+pub async fn create_transaction(
+    State(state): State<AppState>,
+    _perm: RequiresPermission<resource::Credits, operation::CreateAll>,
+    Json(data): Json<CreditTransactionCreate>,
+) -> Result<(StatusCode, Json<CreditTransactionResponse>)> {
+    // Validate that only admin_grant or admin_removal are allowed
+    if !matches!(
```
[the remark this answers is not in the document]
fergusfinn: One very rust-y thing we can do here is make the grant amount a `NonZero`, so that we don't do the runtime validation but fail in the serializer

dwctl/migrations/023_add_credit_system.sql
```diff
@@ -0,0 +1,55 @@
...
+CREATE INDEX idx_credit_transactions_created_at ON credit_transactions (created_at DESC);
+CREATE INDEX idx_credit_transactions_type ON credit_transactions (transaction_type);
+
+-- Comments
+COMMENT ON TABLE credit_transactions IS 'Simple ledger of all credit transactions';
+COMMENT ON COLUMN credit_transactions.transaction_type IS 'Type of transaction - purchase, usage or admin adjustment';
+COMMENT ON COLUMN credit_transactions.amount IS 'Absolute value of transaction amount';
+COMMENT ON COLUMN credit_transactions.balance_after IS 'Balance after this transaction';
+
+-- Make the table append-only (prevent updates and deletes)
```
fergusfinn: ok sure - i think we should only have one write point really? Like i mentioned above - outlet-postgres kinda only calls from dwctl code, and the billing integration will also call from here. But that's a bit irrelevant, i like the extra protection here

dwctl/migrations/023_add_credit_system.sql
```diff
@@ -0,0 +1,55 @@
+-- Add credit system for billing and usage tracking
+
+-- Create transaction type enum
+CREATE TYPE credit_transaction_type AS ENUM (
```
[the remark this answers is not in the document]
fergusfinn:
```
    -- Status tracking
    status TEXT NOT NULL CHECK (status IN ('pending', 'processing', 'completed', 'failed', 'canceled')),
```

You can also do this - i don't know if this works out as more atomic. I'd argue for a plain text field, but i'm happy to defer to your judgement if you think the postgres enum is better and are happy to deal with checking that the migration behaviour is ok, if we need to extend it later

dwctl/src/db/handlers/credits.rs
```diff
@@ -0,0 +1,743 @@
...
+
+impl<'c> Credits<'c> {
+    pub fn new(db: &'c mut PgConnection) -> Self {
+        Self { db }
+    }
+
+    /// Create a new credit transaction
+    /// This method validates the balance_after is correct based on the current balance
+    pub async fn create_transaction(&mut self, request: &CreditTransactionCreateDBRequest) -> Result<CreditTransactionDBResponse> {
+        let mut tx = self.db.begin().await?;
```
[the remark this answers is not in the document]
pjb157: It now works but if you want check out the final two unit tests in credits.rs

PR #53, opened 2025-10-31
Add support for billing to frontend

JoshC8C7:
Currently hidden behind the use_billing feature flag.

Also fixes demo mode to use the MSW for retrieving the current user.

### DIFF
--- a/dashboard/src/components/features/cost-management/CostManagement/TransactionHistory.tsx
+++ b/dashboard/src/components/features/cost-management/CostManagement/TransactionHistory.tsx
@@ -179,12 +179,12 @@ export function TransactionHistory({
       <Card className="p-4">
         <div className="flex items-center justify-between mb-2">
           <h2 className="text-2xl font-semibold text-doubleword-neutral-900">
-            Transactions
+            Transaction History
           </h2>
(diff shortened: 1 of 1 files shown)


dashboard/src/api/control-layer/client.ts
```diff
@@ -699,6 +704,49 @@ const authApi = {
...
+    const url = `/admin/api/v1/credits/transactions${params.toString() ? "?" + params.toString() : ""}`;
+    const response = await fetch(url);
+    if (!response.ok) {
+      throw new Error(`Failed to fetch transactions: ${response.status}`);
+    }
+    return response.json();
+  },
+
+  async addCredits(data: AddCreditsRequest): Promise<AddCreditsResponse> {
+    const response = await fetch("/admin/api/v1/credits/add", {
```
fergusfinn: We didn't settle on this API did we?

fergusfinn: i.e. credits/add is really bad from a REST perspective - in the backend API PR i thought we had POSTing to /transactions for this. Is this path just for MSW?

JoshC8C7:
> Where are we on merging this with a feature flag?

Should just be a matter of fixing tests now.

fergusfinn:
we good to merge?

PR #66, opened 2025-11-06
Peter/dw 123 deduct credits when users use api

pjb157:
Adds:
* pricing to onwards sync which updates onwards response headers added in: https://github.com/doublewordai/onwards/pull/29
* Creates usage transaction when requests are sent via the API
* Only sync api keys if credits are sufficient

Fixes:
* Removes check on balance after to be > 0 and places that on the amount. This was a mistake in the previous PR.


fergusfinn:
This is nice - it comes out really cleanly with the response headers changes

PR #111, opened 2025-11-14
allow pricing to be updated from frontend

JoshC8C7:
- Lets you set balance on frontend
- Adds a trigger so balance crossing zero enables/disables the api key

### DIFF
--- a/dwctl/src/db/handlers/credits.rs
+++ b/dwctl/src/db/handlers/credits.rs
@@ -836,8 +836,7 @@ mod tests {
 
         assert_eq!(notification.channel(), "auth_config_changed");
 
-        let payload: serde_json::Value =
-            serde_json::from_str(notification.payload()).expect("Failed to parse notification payload");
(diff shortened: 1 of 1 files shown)


dwctl/migrations/030_notify_on_balance_threshold.sql
```diff
@@ -0,0 +1,38 @@
+-- Add trigger to notify onwards when user balance crosses the zero threshold
+-- This ensures API keys are removed when balance goes negative and restored when it goes positive
+
+CREATE OR REPLACE FUNCTION notify_on_balance_threshold_crossing() RETURNS trigger AS $$
+DECLARE
+    old_balance DECIMAL(12, 8);
+BEGIN
```
fergusfinn: hmm why do we have to do this? Don't we currently notify on all balance changes? This seems like it would be useful if we were more careful about how we notified

JoshC8C7: there's a test to elucidate this now

PR #151, opened 2025-11-19
Batch validate model access

sejori:


fergusfinn:
Looks good to me! happy to merge when you feel like its ready, up to you how important you reckon my above anxieties are

PR #168, opened 2025-11-19
Fix admin credits

sejori:


dwctl/migrations/031_fix_balance_threshold_decimal_precision.sql
```diff
@@ -0,0 +1,34 @@
+-- Fix decimal precision in balance threshold trigger function
+-- The function was using DECIMAL(12, 8) but should use DECIMAL(12, 2) to match
+-- the column precision updated in migration 024
+
+CREATE OR REPLACE FUNCTION notify_on_balance_threshold_crossing() RETURNS trigger AS $$
+DECLARE
+    old_balance DECIMAL(12, 2);
```
fergusfinn: We should think about precision in the balance column here. OAI (for example) pricing is ~0.05 dollars/million tokens (for input) and 0.400 dollars/million for output. But we're trying to track balance changes at per-request granularity. Minimum for a single request is 1 input & 1 output token - so (0.405 * 10^-6)$. Which will be rounded to zero in the amount column and the balance_after column, so balance will never change. 

Is there any good reason to keep this number small? We can make it like (64, 32) and get a 64 bit decimal number, with 32 decimal points? We definitely shouldn't be making it small for performance reasons, but even if we were its definitely getting packed into a 64 bit int somewhere anyway.

PR #170, opened 2025-11-20
feat: add actual cancellation of in-progress batch requests

fergusfinn:
Prior to this PR, cancellation only affected pending & claimed requests - any requests that were 'in progress' were left to complete by themselves. This PR adds a LISTEN/NOTIFY channel for best-effort cancellation of the actual HTTP requests as well. This should mean we stop processing cancelled requests more quickly. 

The mechanism is also useful because it means we can interrupt HTTP requests when the server shuts down gracefully - for example on SIGTERM. This is useful because we then don't have to rely on the retry mechanism for requests in flight when a pod is stopped gracefully by kubernetes - for example to move nodes, or to trigger an upgrade.

### DIFF
--- a/fusillade/src/manager/postgres.rs
+++ b/fusillade/src/manager/postgres.rs
@@ -2599,82 +2599,97 @@ impl<H: HttpClient + 'static> PostgresRequestManager<H> {
         let pool = self.pool.clone();
 
         // Create a stream that handles reconnection internally
-        let stream = futures::stream::unfold(pool, |pool| async move {
+        // State: (pool, optional listener) - we keep the listener alive across iterations
+        type State = (PgPool, Option<sqlx::postgres::PgListener>);
+
+        let stream = futures::stream::unfold((pool, None), |(pool, listener_opt): State| async move {
             const RECONNECT_DELAY_SECS: u64 = 5;
 
-            'reconnect: loop {
-                // Try to create a new listener
-                let mut listener = match sqlx::postgres::PgListener::connect_with(&pool).await {
-                    Ok(l) => {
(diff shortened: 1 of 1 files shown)


fusillade/src/manager/postgres.rs
```diff
@@ -2569,14 +2574,65 @@ impl<H: HttpClient + 'static> DaemonExecutor<H> for PostgresRequestManager<H> {
...
+                    match notification.payload().parse::<uuid::Uuid>() {
+                        Ok(uuid) => Some((RequestId(uuid), listener)),
+                        Err(e) => {
+                            tracing::error!(
+                                payload = notification.payload(),
+                                error = %e,
+                                "Failed to parse request ID from notification"
+                            );
+                            // Continue listening despite parse error
+                            Some((RequestId(uuid::Uuid::nil()), listener))
```
fergusfinn: This is a bit dodgy maybe - we're using a nil UUID to advance the stream. This relies on the UUID of a valid request never being nil, which is definitely true atm, since they're database generated.

sejori: I think using the nil UUIDs is perfectly valid - a NULL case could be added or you could wrap it in a loop (claude suggestion) but it's marginal. 

The design of controlling a stream that the daemon listens to is super nice and likely the most reactive way to provide cancellation. This is a push-only subscription which I guess means daemons should be scaled by their maximum anticipated simultaneous cancellation events? To make sure we don't overwhelm when cancelling thousands of requests at once? I'm not sure if this is a really an issue. What do you think?

One minor point would be that this utility feels very central to the core postgres pub/sub system so it certainly belongs here but it feels quite buried amongst other logic right now.

PR #174, opened 2025-11-20
feat: improve the usability of the proxy-header configuration

fergusfinn:
Currently proxy-header auth only accepts one header, that contains the email address of the user to be logged in. This plays two separate roles: 
1. It's the user's email (obviously)
2. It's the unique identifier that powers 'get_or_create_user', which is called in the `current_user.rs` file. This means that any two users with the same email at some upstream proxy are treated as being the same.

This is generally fine, when emails are unique. That's not a great assumption though. For example, Auth0/AWS Cognito, etc. as a provider plugged into something like [oauth2-proxy](https://oauth2-proxy.github.io/oauth2-proxy/) allow you to setup multiple providers (for example, google, github, database). Users can create accounts at any of these providers, and logins from the same user can come from different places. This opens up a risk for a user takeover attack - if an upstream provider doesn't verify emails, then an attacker can signup at a provider with a known email and access the data for a user that signed in with another method. It also causes problems when user's change their emails at upstream providers - we'll create a new user when they login after the email change, instead of following them to their new email.

This PR introduces a new external_user_id field to uniquely identify users, separate from their email address, and a new header, 'X-Doubleword-Email' (configurable) in which the email is supplied (X-Doubleword-User should be used for the unique identifier. Here's the new config: 

```yaml
  proxy_header:
    enabled: false
    # header_name: User identifier or email
    # Single header mode: User's email (e.g., "user@example.com")
    # Dual header mode: Unique identifier from IdP (e.g., "github|user123", "google-oauth2|456")
    header_name: "x-doubleword-user"
    # email_header_name: User's email address (optional, enables dual header mode)
    # If provided: Enables federated identity with (email, external_user_id) uniqueness
    # If omitted: Uses header_name value as email (single header mode, email must be unique)
    email_header_name: "x-doubleword-email"
```

## Database Changes:
- New external_user_id column on users table (nullable, with unique index)
- Migration adds column but does not backfill, values populate on first login

##  Configuration Changes:

- proxy_header.header_name (default: x-doubleword-user) - Contains the unique user identifier from the IdP (e.g.,
  auth0|google-oauth2|123456)
- proxy_header.email_header_name (default: x-doubleword-email) - Contains the user's email address
- if email_header name isn't supplied, we fall back to assuming that header name contains an email address that can be used for both email and identifier.

##  For existing deployments

1. Deploy new dwctl code with migration
2. Existing users have external_user_id = NULL initially
3. Update proxy configuration to send both headers
4. On first login after deployment, their external_user_id is backfilled from the X-Doubleword-User header
5. Subsequent logins use external_user_id directly

### DIFF
--- a/dwctl/src/db/handlers/users.rs
+++ b/dwctl/src/db/handlers/users.rs
@@ -12,8 +12,6 @@ use crate::{
 use chrono::{DateTime, Utc};
 use serde::{Deserialize, Serialize};
 use sqlx::{Connection, FromRow, PgConnection};
-use std::collections::hash_map::DefaultHasher;
-use std::hash::{Hash, Hasher};
 use tracing::instrument;
 use uuid::Uuid;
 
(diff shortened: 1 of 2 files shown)


dwctl/src/db/handlers/users.rs
```diff
@@ -345,6 +354,209 @@ impl<'c> Users<'c> {
...
+                        group_repo
+                            .sync_groups_with_sso(
+                                user.id,
+                                groups,
+                                provider,
+                                &format!("A group provisioned by the {provider} SSO source."),
+                            )
+                            .await?;
+                    }
+                    return Ok(user);
```
sejori: Each path seems to call `sync_groups_with_sso` before returning `OK(User)`. Potentially there is a way to flatten this control flow and perform this once at the end to make a bit DRYer/reduce the line count?

PR #215, opened 2025-11-25
refactor: optimize onwards_config with single mega-query and add E2E test

fergusfinn:
Refactor onwards_config.rs to use a single optimized query instead of multiple round-trips to the database. This significantly improves performance when syncing configuration updates.

**Refactoring changes:**
- Replace multiple sequential queries with one mega-query using JOINs and LATERAL
- Add `OnwardsTarget` and `OnwardsApiKey` structs for cleaner data handling
- Make `load_targets_from_db()` public for testing
- Add `sync_onwards_config()` method to `BackgroundServices` for manual sync triggering
- Query now handles all access control logic in a single database round-trip

**Performance improvements:**
- Reduced from ~4+ queries per sync to 1 query
- Better utilization of PostgreSQL query planner
- Cleaner separation of concerns with dedicated structs

**New E2E integration test:**
- `test_e2e_ai_proxy_with_mocked_inference`: Full AI proxy flow test (~300 lines)
- Tests complete user journey:
  1. Admin creates inference endpoint via API
  2. Admin creates deployment/model via API
  3. Admin creates group and assigns deployment via API
  4. Admin adds user to group via API
  5. User creates API key via API
  6. Manual sync of onwards config (tests the new sync method)
  7. User makes inference request through AI proxy
  8. Validates response and request logging
- Uses wiremock to mock inference endpoint responses
- Validates routing cache updates work correctly

**Documentation:**
- Add CLAUDE.md guidance about avoiding `tokio::time::sleep` in tests
- Recommends polling patterns for faster, more robust tests

**Dependencies:**
- Based on #211 (Rust Edition 2024) - requires Edition 2024 for let-chains used in refactored code

### DIFF
--- a/dwctl/src/lib.rs
+++ b/dwctl/src/lib.rs
@@ -268,10 +268,18 @@ pub fn migrator() -> sqlx::migrate::Migrator {
 /// # }
 /// ```
 #[instrument(skip_all)]
-pub async fn create_initial_admin_user(email: &str, password: Option<&str>, db: &PgPool) -> Result<UserId, sqlx::Error> {
+pub async fn create_initial_admin_user(
+    email: &str,
+    password: Option<&str>,
+    argon2_params: password::Argon2Params,
+    db: &PgPool,
+) -> Result<UserId, sqlx::Error> {
     // Hash password if provided
     let password_hash = if let Some(pwd) = password {
-        Some(password::hash_string(pwd).map_err(|e| sqlx::Error::Encode(format!("Failed to hash admin password: {e}").into()))?)
+        Some(
+            password::hash_string_with_params(pwd, Some(argon2_params))
+                .map_err(|e| sqlx::Error::Encode(format!("Failed to hash admin password: {e}").into()))?,
+        )
     } else {
         None
     };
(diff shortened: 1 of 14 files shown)


dwctl/src/sync/onwards_config.rs
```diff
@@ -202,88 +233,199 @@ impl OnwardsConfigSync {
...
+    )
+    .fetch_all(db)
+    .await?;
 
-        // Fetch all deployments
-        models = deployments_repo.list(&DeploymentFilter::new(0, i64::MAX)).await?;
-    }
+    let query_duration = query_start.elapsed();
+    info!(
+        "Mega-query completed in {:?}, fetched {} rows ({} rows/ms)",
```
pjb157: Haha are we officially referring to it as mega query

dwctl/src/lib.rs
```diff
@@ -947,6 +949,27 @@ impl BackgroundServices {
...
+        // Use the same load function as the automatic sync
+        let new_targets = crate::sync::onwards_config::load_targets_from_db(pool).await?;
+
+        // Send through the watch channel (same as automatic sync)
+        sender
+            .send(new_targets)
+            .map_err(|_| anyhow::anyhow!("Failed to send targets update"))?;
+
+        Ok(())
+    }
```
sejori: Could there be a potential need for this as an admin? A force trigger from the UI in case a notify event seems flaky?

fergusfinn: yeah its a good point - in principle LISTEN/NOTIFY is 'at most once' - so if you're not listening (i.e. a temporary db disconnect) you don't get the event. What i was planning to intead at some point was to add a 'fallback sync', where every second, if an event hasn't been received, we trigger a sync anyway. That way we have a bound on how inconsistent the cache can be?

dwctl/src/sync/onwards_config.rs
```diff
@@ -22,13 +21,41 @@ pub enum SyncStatus {
...
+
+/// Minimal API key data needed for onwards config
+#[derive(Debug, Clone)]
+struct OnwardsApiKey {
+    id: ApiKeyId,
+    secret: String,
+    requests_per_second: Option<f32>,
+    burst_size: Option<i32>,
+}
+
```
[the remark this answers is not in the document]
fergusfinn: there's nothing really analogous over there unfortunately - theses are plain data structs to hold data from the query, and those are more applicatoin based (holding actual rate limitiers etc.)

PR #218, opened 2025-11-25
UX: Billing UX improvements

fergusfinn:
Lots of little UI/UX tweaks: 

4. Move the billing page into the 'user' dropdown, and call it 'billing'. 
5. Get rid of the three dots click on the per-user page, and instead make the button redirect to the transaction history page, with a query flag set. 
6. Make the add funds button not blue.


JoshC8C7:
Can we merge this after #178 - i'm currently perpetually merging and this one has a fair bit of overlap.

PR #308, opened 2025-12-02
fix: cancel safety in background services

fergusfinn:
This should fix a bug we where, when pressing CTRL-C on a running control layer instance, the server would shutdown, but the background tasks would continue to run to completion. This meant that the batch daemon would exhaust the retries for any running requests, since they would hit the (now shutdown) control layer. 

The problem was in this code: 

```rust
        let server_error: Option<anyhow::Error> = tokio::select! {
            result = axum::serve(listener, service.into_make_service()).with_graceful_shutdown(shutdown) => {
                result.err().map(Into::into) // None if server shut down cleanly
            }
            result = self.bg_services.wait_for_failure() => {
                // Background task failed - save error for fail-fast restart after cleanup
                match result {
                    Ok(_infallible) => unreachable!("wait_for_failure never returns Ok"),
                    Err(e) => Some(e),
                }
            }
        };
```

If the first branch of the select is taken (which happens if CTRL-C is pressed), the second future is cancelled. Then, we call
```rust
        // Graceful shutdown - even if we're failing fast, clean up properly
        info!("Shutting down background services...");
        self.bg_services.shutdown().await;
```
to shutdown the remaining background services. The problem is that `self.bg_services.wait_for_failure()` isn't [cancel-safe](https://docs.rs/tokio/latest/tokio/macro.select.html#cancellation-safety). Internally, it takes all the tasks out of the internal buffer, and stores them in a local variable, and tries to join them all. If, in the meantime, this join is cancelled, the local variable gets dropped, and the tasks are no longer in the buffer, and so the shutdown method doesn't wait for them, and instead proceeds to closing the pool and waiting for all its connections to be returned. 

The solution is to replace this with a `JoinSet`, which is the tokio native way to do this thing of spawning multiple tasks and keeping them together in a collection.

This doesn't fix the whole issue though. I still don't understand why the cancellation token in the shutdown method isn't shutting down the (now detached) tasks.


sejori:
LGTM - let's try it 🤓

PR #372, opened 2025-12-15
support grouping transactions by batch on backend

JoshC8C7:
Amalgamates individual transactions that forms a bitch into one, so that it can be presented on the frontend.


dwctl/src/db/handlers/credits.rs
```diff
@@ -248,6 +257,188 @@ impl<'c> Credits<'c> {
...
+    #[instrument(skip(self), fields(user_id = ?user_id.map(|id| abbrev_uuid(&id)), skip = skip, limit = limit), err)]
+    pub async fn list_transactions_with_batches(
+        &mut self,
+        user_id: Option<UserId>,
+        skip: i64,
+        limit: i64,
+    ) -> Result<Vec<(CreditTransactionDBResponse, Option<Uuid>)>> {
+        let rows = sqlx::query!(
+            r#"
+            SELECT * FROM (
```
fergusfinn: oh boy this query scares me. How fast is it on prod? have we got the right indices on it?

JoshC8C7: the join of transactions with http_analytics is kind of inavoidable here - its done twice technically but only because its 'blocked' by splitting the batch and non batch usages. But yeah I appreciate it could be slow.

PR #445, opened 2026-01-06
fix: compute balance on read instead of on write

fergusfinn:


dwctl/src/api/handlers/transactions.rs
```diff
@@ -200,26 +200,58 @@ pub async fn list_transactions(
...
-        let result: Vec<CreditTransactionResponse> = transactions_with_batches
+        let count = repo.count_transactions_with_batches(filter_user_id).await?;
+        let txs: Vec<CreditTransactionResponse> = transactions_with_batches
             .into_iter()
             .map(|(tx, batch_id)| CreditTransactionResponse::from_db_with_batch_id(tx, batch_id))
             .collect();
+        (txs, count)
+    } else if let Some(user_id) = filter_user_id {
+        let txs = repo.list_user_transactions(user_id, skip, limit).await?;
+        let count = repo.count_user_transactions(user_id).await?;
```
fergusfinn: why do we count with a separate db request here? Seems like we could just count the txs

dwctl/src/db/handlers/credits.rs
```diff
@@ -8,13 +8,19 @@ use crate::{
...
-use sqlx::{Connection, FromRow, PgConnection};
+use sqlx::{FromRow, PgConnection};
 use std::collections::HashMap;
 use tracing::{error, instrument, trace};
 use uuid::Uuid;
 
+/// Probability of refreshing checkpoint on each transaction (1 in N).
+/// With N=1000, checkpoint lags by ~1000 transactions on average,
+/// meaning balance reads aggregate ~500 rows on average.
+const CHECKPOINT_REFRESH_PROBABILITY: u32 = 1000;
```
fergusfinn: This is an interesting thing to do - i made it probabilistic because then i didn't have to think about coordination between replicas.

PR #517, opened 2026-01-13
show sla on model summaries

JoshC8C7:
Before and after:

image

image


sejori (commented):
I like it - clean and concise. Since user's are expecting batch pricing this should make total sense.

PR #662, opened 2026-02-05
fix: stream batch results and file content to prevent OOM

fergusfinn:
## Summary

- **`get_batch_results`** and **`get_file_content`** were collecting all results into memory (`.collect().await`) before sending the response. For large result sets (e.g., 50k embeddings at ~16KB each ≈ 800MB), this caused OOM kills on pods with 4Gi memory limits.
- For **paginated requests** (with `limit`): buffer only N+1 items to detect if more pages exist
- For **unlimited requests**: true streaming via `Body::from_stream()`, never holding more than one item in memory
- Also fixes `get_batch_results` which was missing the "still processing" check for `X-Incomplete` header (`get_file_content` already had this correctly)

## Test plan

- [x] All 732 existing tests pass
- [ ] Verify paginated requests still return correct `X-Incomplete` and `X-Last-Line` headers
- [ ] Verify unlimited downloads stream correctly (e.g., large embedding batch results)
- [ ] Verify `X-Incomplete` is `true` when batch is still processing

### DIFF
--- a/dwctl/src/api/handlers/files.rs
+++ b/dwctl/src/api/handlers/files.rs
@@ -27,7 +27,7 @@ use crate::types::Resource;
 use axum::{
     Json,
     body::Body,
-    extract::{Multipart, Path, Query, State},
+    extract::{FromRequest, Multipart, Path, Query, State},
     http::StatusCode,
 };
 use bytes::Bytes;


sejori (commented):
Looks good to me 👍

PR #676, opened 2026-02-10
feat: add rate limiting and fallback sync for onwards notifications

pjb157:
## Summary

Mitigates notification storm incidents by adding global rate limiting and a fallback sync mechanism to the onwards sync system.

### Rate Limiting
- Replace per-user rate limiting with global rate limiting in analytics batcher
- When users have depleted balances, send one notification that triggers full cache reload (not one per user)
- Configurable via `analytics.balance_notification_interval_seconds` (default: 5s)
- Add metrics: `dwctl_onwards_sync_notifications_total{action="sent|rate_limited"}`

### Fallback Sync
- Add periodic fallback timer to onwards sync (default: 10s)
- Ensures eventual consistency even if LISTEN/NOTIFY fails
- Independent of notification system - runs on timer
- Add metrics: `dwctl_cache_sync_total{source="fallback"}`

### Documentation
- Add `docs/onwards-sync-notification-system.md` explaining:
  - System architecture (LISTEN/NOTIFY + fallback timer)
  - Configuration options
  - Multi-instance behavior
  - Monitoring and troubleshooting

## Configuration

```yaml
analytics:
  balance_notification_interval_seconds: 5

background_services:
  onwards_sync:
    fallback_interval_seconds: 10
```

## Design Rationale

**Global rate limiting**: The onwards sync reloads ALL user data, so per-user rate limiting sends redundant notifications. One notification is sufficient.

**Fallback timer**: LISTEN/NOTIFY can miss events (connection issues, notifications lost). Fallback ensures cache syncs eventually with bounded staleness.

**In-memory rate limiting**: Avoids database queries in the hot path (analytics batcher). Works well in multi-instance deployments (each instance rate-limits independently).

## Testing

- [ ] Add unit tests for global rate limiter in `batcher.rs`
- [ ] Add integration test for fallback sync behavior
- [ ] Add test for notification storm prevention
- [ ] Add test for multi-instance rate limiting behavior


dwctl/src/sync/onwards_config.rs
```diff
@@ -156,6 +176,15 @@ impl OnwardsConfigSync {
...
             }
             info!("Started onwards configuration listener");
+
+            // Create fallback sync timer (if enabled)
+            let mut fallback_timer = fallback_interval.map(|interval| {
+                let mut timer = tokio::time::interval(interval);
+                // Use Delay to avoid burst of syncs after runtime hiccups
+                timer.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Delay);
+                timer
+            });
```
sejori: So if we set a fallback interval of zero would this mean the timer is zero so we manually sync every tick? That sounds like an overload of db connections etc. 

The comment above says 0=disabled so maybe I am wrong - if we have a value of 0 do we not perform background syncs?

sejori (commented):
LGTM - one commen, not sure if it's a legit issue

dwctl/src/sync/onwards_config.rs
pjb157: Line 161-165 in onwards_config.rs:
```rust                                                                                                                                                                            
// Fallback sync interval (0 = disabled)                                                                                                                                                                      
let fallback_interval = if config.fallback_interval_seconds > 0 {                                                                                                                                             
    Some(std::time::Duration::from_secs(config.fallback_interval_seconds))                                                                                                                                    
} else {                                                                                                                                                                                                      
    None  // Returns None when 0                                                                                                                                                                              
};
```

PR #724, opened 2026-02-17
feat: add trusted provider flag for strict mode response sanitization

pjb157:
## Summary

Adds support for marking model providers as trusted when strict mode is enabled. Trusted providers bypass error response sanitization, while non-trusted providers have sensitive error details removed.

## Configuration

**Enable Strict Mode**

In `config.yaml`:
```yaml
onwards:
  strict_mode: true
```

Or via environment variable:
```bash
DWCTL_ONWARDS__STRICT_MODE=true
```

When strict mode is enabled:
- **Trusted providers** (`trusted: true`): Error responses pass through unchanged, showing full error details
- **Non-trusted providers** (`trusted: false`, default): Error responses are sanitized to remove sensitive information (internal IPs, stack traces, database details)

## Changes

- Add `trusted` boolean column to `deployed_models` table (migration 066, defaults to `false`)
- Expose `trusted` field in model creation and update APIs
- Integrate trusted flag with onwards strict mode sanitization logic
- Add comprehensive test coverage for strict mode behavior
- Add `onwards.strict_mode` to config response API

## API Usage

Create a trusted provider:
```json
POST /admin/api/v1/models
{
  "model_name": "gpt-4",
  "trusted": true
}
```

Update existing provider:
```json
PATCH /admin/api/v1/models/{id}
{
  "trusted": true
}
```

Check strict mode status:
```json
GET /admin/api/v1/config

Response:
{
  "onwards": {
    "strict_mode": true
  },
  ...
}
```

## Testing

- All 884 tests passing
- Clippy clean
- All Copilot PR review comments addressed

## Notes

Backend and frontend implementations are complete and production-ready.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,16 +1,14 @@
 name: CI
 
 on:
-  push:
-    branches: [main]
   pull_request:
     branches: [main]
-  release:
-    types: [published, edited]
+  merge_group:
   workflow_dispatch:
 
 jobs:
   frontend-test:
+    if: (github.event_name == 'pull_request' && !startsWith(github.head_ref, 'release-please--')) || github.event_name == 'merge_group'
     runs-on: depot-ubuntu-24.04-8
 
     steps:
@@ -95,6 +93,7 @@ jobs:
           curl -X POST "https://charts.somnial.co/doubleword-control-layer/gzipped-bundle-size?value=${GZIPPED_SIZE}" || true
 
   backend-test:
+    if: (github.event_name == 'pull_request' && !startsWith(github.head_ref, 'release-please--')) || github.event_name == 'merge_group'
     runs-on: depot-ubuntu-24.04-16
 
     services:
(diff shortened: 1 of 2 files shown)


dwctl/src/api/models/deployments/enrichment.rs
```diff
@@ -442,6 +442,7 @@ mod tests {
             fallback: None,
             components: None,
             sanitize_responses: None,
+            trusted: None,
```
hachall: i think we should sanitise this information to standard users? we hide things like rate limits and composite info from them, like so:

```rust
// Mask rate limiting info for users without ModelRateLimits permission
            if !self.can_read_rate_limits {
                model_response = model_response.mask_rate_limiting();
                model_response = model_response.mask_capacity();
            }

            if !self.can_read_users {
                model_response = model_response.mask_created_by();
            }

            // Mask composite model info for users without permission
            if !self.can_read_composite_info {
                model_response = model_response.mask_composite_fields();
            }
```

hachall:
in the UI for the virtual model i can see how to toggle santize responses, but how do i set 'trusted'?

PR #728, opened 2026-02-18
fix: advance to step 2 when manually configuring models in edit endpoint modal

### LINKED ISSUE #550: Unable to Edit Models in Existing Endpoint
Proxying through `tensorzero` behind `control-layer` i'm finding that i'm unable to edit model configurations for the endpoint. If they change on the back-end, they can't be updated. TZ doesn't respond to `/models` apparently but the manual path simply does nothing when clicking `next`. Have tried adding `/` on the end of the URI to prompt a "content changed" unclogging to no avail.

fergusfinn:
## Summary

- Fixes `handleConfigureManually` in `EditEndpointModal` to call `setCurrentStep(2)`, matching the behavior in `CreateEndpointModal`
- Without this, clicking "Next" with auto-discover disabled left users stuck on step 1 with no way to edit models on an existing endpoint
- One-line fix — the rest of the manual mode flow (model input, "Configure Models" button, alias configuration) already works correctly once step 2 is reached

## Root cause

`EditEndpointModal.handleConfigureManually` (line 287) was missing `setCurrentStep(2)`. The equivalent function in `CreateEndpointModal` (line 445) includes this call. This meant that when auto-discover was disabled (necessary for non-OpenAI backends like TensorZero that don't serve `/v1/models`), the "Next" button toggled manual mode but never advanced the wizard step.

## Test plan

- [ ] Open an existing endpoint for editing
- [ ] Uncheck "Auto-discover models"
- [ ] Click "Next" — should now advance to step 2 with the manual model input textarea
- [ ] Enter model names, click "Configure Models" — model selection/alias UI should appear
- [ ] Complete the edit and save — endpoint should update with new model configuration
- [ ] Verify creating a new endpoint still works as before (no regression)

Fixes #550

### DIFF
--- a/dashboard/src/components/modals/EditEndpointModal/EditEndpointModal.test.tsx
+++ b/dashboard/src/components/modals/EditEndpointModal/EditEndpointModal.test.tsx
@@ -124,7 +124,7 @@ describe("EditEndpointModal", () => {
     });
   });
 
-  it("shows API key field when endpoint requires API key", () => {
+  it("shows API key field with hint when endpoint requires API key", () => {
     render(
       <EditEndpointModal
         isOpen={true}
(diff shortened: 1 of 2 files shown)


JoshC8C7 (approved):
if its been tested manually then sounds good

PR #729, opened 2026-02-18
feat: User usage dashboard

JoshC8C7:
## Summary                                                                                                                                                                                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                                                                                                                                                         
  Adds a new Usage page showing per-user batch usage metrics with date filtering.                                                                                                                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                                                                                                                                                       
  ### Frontend                                                                                                                                                                                                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                                                                                                                                                                         
  New `/usage` route accessible to all roles. The page shows:
  - Total input/output tokens, request count, batch count
  - Per-model breakdown table with tokens, cost, and request count
  - Estimated realtime cost (what the same usage would cost at realtime rates)
  - Date range picker (defaults to last 30 days, can be cleared for all-time view)

  Changes:
  - `Usage.tsx` — new page component with `DateTimeRangeSelector`
  - `App.tsx` — route registration
  - `AppSidebar.tsx` — nav entry (Activity icon)
  - `authorization.ts` — `usage` permission added to all roles
  - `types.ts` — `UserBatchUsageResponse` and `ModelBreakdownEntry` types
  - `client.ts` — `usageApi.get()` with optional date params
  - `hooks.ts` — `useUsage(startDate?, endDate?)` hook
  - `keys.ts` — `usage.forRange()` query key for cache invalidation

  ### Backend

  `GET /admin/api/v1/usage` with optional `start_date`/`end_date` query params. Two paths depending on whether dates are provided:

  **With dates** (capped at 30 days) — queries `http_analytics` directly via covering index:

  get_user_model_breakdown_for_range()  →  per-model tokens/cost/requests from http_analytics
  get_user_batch_count_for_range()      →  COUNT(DISTINCT fusillade_batch_id) from http_analytics
  get_realtime_tariffs()                →  ~12 rows from model_tariffs
  → realtime cost computed in Rust: sum(tokens × tariff) per model

  **Without dates** — all-time stats from pre-aggregated tables (60s cache):

  refresh_user_model_usage()            →  incremental cursor upsert of new rows
  get_user_model_breakdown()            →  per-model breakdown from user_model_usage table
  get_user_batch_counts()               →  batch count/avg from batch_aggregates (pre-existing)
  get_realtime_tariffs()                →  ~12 rows from model_tariffs
  → realtime cost computed in Rust: sum(tokens × tariff) per model

  Both paths share response assembly and the tariff-based realtime cost calculation. All queries filter on `fusillade_batch_id IS NOT NULL` — only batched requests are included.

  ### Queries

  | Function | Source | Used by |
  |----------|--------|---------|
  | `get_user_model_breakdown` | `user_model_usage` table | All-time |
  | `get_user_batch_counts` | `batch_aggregates` table (pre-existing) | All-time |
  | `get_user_model_breakdown_for_range` | `http_analytics` via covering index | Date-filtered |
  | `get_user_batch_count_for_range` | `http_analytics` via covering index | Date-filtered |
  | `get_realtime_tariffs` | `model_tariffs` + `deployed_models` (~12 rows) | Both |
  | `refresh_user_model_usage` | Incremental cursor upsert | All-time |

  ### Migrations

  **066 — `user_model_usage` summary table**

  Pre-aggregated per-user per-model usage. Turns an O(N) scan of `http_analytics` (~14M rows) into an O(M) lookup where M = distinct models per user (~3). Maintained incrementally via a cursor tracking `last_processed_id` so only new rows are scanned on each refresh. Backfills existing data on creation.

  **067 — Covering index `idx_analytics_user_usage`**

  Replaces `idx_analytics_user_id` (same key columns, 884 MB, 0 lifetime scans). The old index was never used because `user_id` has poor heap correlation (-0.24) — user rows are scattered across the table, so fetching from the heap meant random I/O slower than scanning by timestamp. The new index adds `INCLUDE (model, prompt_tokens, completion_tokens, total_cost, fusillade_batch_id)` so queries are
  answered entirely from the index (Index Only Scan). 1,702 MB (+818 MB net).

  Measured on heaviest user (11% of traffic, 1.5M rows):

  | Window | Before | After |
  |--------|--------|-------|
  | 30d | ~35s | 53ms |
  | 7d | 864ms | 11ms |
  | 1d | 0.4ms | 0.1ms |


[Screencast from 2026-02-19 18-13-32.webm](https://github.com/user-attachments/assets/2bb5600f-af00-4300-9b6f-375e54c599b6)

### DIFF
--- a/dashboard/src/components/features/usage/Usage.tsx
+++ b/dashboard/src/components/features/usage/Usage.tsx
@@ -98,13 +98,17 @@ function DonutChart({ data, size = 190, thickness = 24, centerLabel }: { data: D
     const angle = total > 0 ? (d.value / total) * 360 : 0;
     const s = cum;
     cum += angle;
+    const fullCircle = angle >= 359.99;
     const sRad = (s * Math.PI) / 180;
     const eRad = (cum * Math.PI) / 180;
     const large = angle > 180 ? 1 : 0;
     const gap = data.length > 1 ? 0.015 : 0;
     const sG = sRad + gap, eG = eRad - gap;
     return {
-      path: `M ${cx + r * Math.cos(sG)} ${cy + r * Math.sin(sG)} A ${r} ${r} 0 ${large} 1 ${cx + r * Math.cos(eG)} ${cy + r * Math.sin(eG)}`,
+      fullCircle,
+      path: fullCircle
+        ? ""
+        : `M ${cx + r * Math.cos(sG)} ${cy + r * Math.sin(sG)} A ${r} ${r} 0 ${large} 1 ${cx + r * Math.cos(eG)} ${cy + r * Math.sin(eG)}`,
       color: PALETTE[i % PALETTE.length],
       ...d,
     };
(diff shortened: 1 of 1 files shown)


dwctl/src/lib.rs
```diff
@@ -1093,6 +1093,7 @@ pub async fn build_router(
         .route("/requests", get(api::handlers::requests::list_requests))
         .route("/requests/aggregate", get(api::handlers::requests::aggregate_requests))
         .route("/requests/aggregate-by-user", get(api::handlers::requests::aggregate_by_user))
+        .route("/usage", get(api::handlers::requests::get_usage))
```
fergusfinn: eh i could probably quibble with this REST-ishly? we have `/usage` and `/monitoring` and `/requests/aggregate`, all of which roughly do the same sorts of things. Idk if theres a good solution though. If the existing pattern is `/requests/<verb>`, then could this just be the per-user `/results/aggregate`? i don't mind too much, we can just leave it like this

PR #730, opened 2026-02-18
fix: omit Secure attribute from cookies when cookie_secure is false

fergusfinn:
## Summary

- When `cookie_secure` is set to `false`, the `Set-Cookie` header was emitting `Secure=false`. Per the [HTTP spec](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Set-Cookie#secure), `Secure` is a boolean flag — its mere presence causes browsers to treat the cookie as secure-only, regardless of value. `Secure=false` is not valid syntax and browsers interpret it the same as `Secure`, preventing the cookie from being set over plain HTTP.
- Also fixes the logout handler which hardcoded `; Secure` regardless of the `cookie_secure` config setting.

## Test plan

- Set `cookie_secure: false` in `config.yaml`
- Login and verify the `Set-Cookie` response header does **not** contain `Secure`
- Set `cookie_secure: true` and verify `; Secure` is present
- Verify logout cookie also respects the setting


sejori (commented):
LGTM!

PR #987, opened 2026-04-14
feat: monitoring demand endpoint decoupled from completion windows

fergusfinn:
## Summary

Adds `GET /admin/api/v1/monitoring/demand?window=<...>` — same per-model deadline-bucketed counts that the existing `/pending-request-counts` returns, but the caller picks the window shape via query parameter instead of implicitly inheriting `config.batches.allowed_completion_windows`.

Each `window=` entry is either:
- `<end>` — shorthand for `0s:<end>`, matches the legacy "due within N" behaviour. Example: `window=1h,24h`.
- `<start>:<end>` — disjoint range. Example: `window=0s:1h,1h:24h`.

Both bounds are offsets from `now`. Shapes can be mixed in a single call. The raw input string is echoed back as the response key, so scouter-style callers that send `window=1h,24h` still see `"1h"` / `"24h"` keys.

Also supports `service_tiers=` with identical semantics to `/pending-request-counts` (default batch-only; priority-decay top-up applies when `flex` is included), so scouter's `none,flex` usage carries over.

## Why

Scouter reads this endpoint to plan replica allocation and only cares about "demand in the next N hours". The old endpoint picked its buckets from whatever completion-window SLAs the batch API exposed to users — e.g. if product adds a `6h` SLA tier for a customer, scouter's buckets would silently shift under it. This PR treats that coupling as the implementation detail it is and lets planning consumers name their own windows.

It also sets up the next step for realtime-demand-aware scaling: the window grammar extends naturally to negative offsets (`window=-1h` = arrivals in the past hour) in a follow-up, giving scouter a trailing-arrivals signal for realtime models.

## History

Rebuilt on current main after the fusillade monorepo merge: the start/end storage signature (formerly fusillade#217) and the `batches.rs` call-site migration are already on main, so this PR is now just the handler, route, openapi entry, and tests. The earlier crates.io pin/bump choreography is obsolete — fusillade is a path dependency now.

## Kept untouched

`GET /admin/api/v1/monitoring/pending-request-counts` is unchanged — no scouter-side deployment coordination needed. We can remove it once no caller uses it.

## Test plan

- [x] 7 `/demand` tests: permission, missing/empty/malformed window, arbitrary window shapes, zero start, service_tiers.
- [x] Existing `/pending-request-counts` tests untouched and passing (20/20 in the queue module).
- [x] `just lint rust` passes.

## Post-merge

1. Release control-layer, roll to production.
2. Ship the scouter PR switching to `/demand?window=1h,24h` (identical semantics to today).
3. (later) Remove `/pending-request-counts` once nothing reads it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -156,7 +156,7 @@ version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
 dependencies = [
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -167,7 +167,7 @@ checksum = "291e6a250ff86cd4a820112fb8898808a366d8f9f58ce16d1f538353ad55747d"
 dependencies = [
  "anstyle",
  "once_cell_polyfill",
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -2069,7 +2069,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "39cab71617ae0d63f51a36d69f866391735b51691dbda63cf6f96d042b63efeb"
 dependencies = [
  "libc",
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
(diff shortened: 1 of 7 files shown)


fergusfinn:
Review feedback addressed in the follow-up commit:

- **P0 (local fusillade patch)**: already removed in the previous commit — `dwctl/Cargo.toml` now pins `fusillade = "16.0.0"` (on crates.io) and the `[patch.crates-io]` block is gone.
- **P1 (0s coerced to 24h)**: introduced a local strict parser `parse_demand_duration` that returns `None` on malformed/negative input and accepts `0`. Demand's `parse_demand_window` now propagates parse failures, and the handler maps them to 400. `window=0s:1h` now parses `0s` as zero seconds.
- **P2 (malformed input silently accepted)**: `window=foo`, `window=1x`, `window=1h,bad`, `window=-1h:1h` all 400 now. Lenient `parse_window_to_seconds` (batch-API forgiving-on-purpose) is untouched.
- **Test coverage**: added `test_demand_accepts_zero_start` and `test_demand_rejects_malformed_window`.

PR #1003, opened 2026-04-20
docs: add data retention delete policy proposal

pjb157:
## Summary

This PR moves the retention proposal into an initial implementation for batch-facing artefacts.

It adds a configurable retention policy for artefacts we expect to clean up after a reasonable period, while leaving durable product data such as `http_analytics` and user records alone.

## What Changed

- Added `background_services.retention` config with:
  - `enabled`
  - `sweep_interval`
  - `batch_size`
  - `batch_artifacts_default_ttl`
- Applied a default file TTL to batch-related uploads when the caller does not explicitly provide `expires_after`.
- Applied the same default TTL to files created by the sync pipeline.
- Stamped batch-linked artefacts with an internal retention TTL marker so grouped clean-up can happen off batch lifecycle rather than per-table ad hoc rules.
- Added a leader-scoped retention sweeper that:
  - finds terminal batches whose retention window has elapsed
  - soft-deletes those batches
  - soft-deletes linked files
  - deletes standalone expired files that are no longer referenced by active batches
  - calls Fusillade orphan purging to hard-delete `requests` and `request_templates`
- Hid the internal retention metadata from batch API responses.
- Added config parsing tests and retention sweep tests.

## Intended Policy Shape

- Batch request artefacts should be cleaned up as a group.
- `requests`, `request_templates`, and Fusillade file artefacts are the first retention-managed set.
- Batch retention is configured separately from other data classes.
- `http_analytics`, user data, and other durable business records are intentionally out of scope here.

## Config Example

```yaml
background_services:
  retention:
    enabled: true
    sweep_interval: 10m
    batch_size: 100
    batch_artifacts_default_ttl: 7d
```

`batch_artifacts_default_ttl: null` means the control layer does not inject a default TTL.

## Notes

- Explicit `expires_after[...]` values from the OpenAI-compatible Files API still win when supplied.
- The retention TTL stored on batches is internal metadata and is stripped from user-facing responses.
- The sweeper only runs on the leader path, matching the rest of the batch/background lifecycle.

## Verification

- `cargo check -p dwctl`
- `cargo fmt --all`

`cargo test -p dwctl retention` is currently blocked by the repo's SQLx test setup because unrelated query macros require either `DATABASE_URL` or an up-to-date prepared query cache.

### DIFF
--- a/docs/data-retention-delete-policy.md
+++ b/docs/data-retention-delete-policy.md
@@ -2,11 +2,11 @@
 
 ## Goal
 
-Introduce a clear retention policy for user-facing artefacts so that soft deletes become real data cleanup after an optional time-to-live, without treating all data the same.
+Introduce a clear retention policy for user-facing artifacts so that soft deletes become real data cleanup after an optional time-to-live, without treating all data the same.
 
 The main intent is:
 
-- Clean up processing artefacts that should not live forever.
+- Clean up processing artifacts that should not live forever.
 - Keep analytics and durable account data out of this policy by default.
 - Make retention understandable as a product and operations concept, not just a table-by-table implementation detail.
 
@@ -25,17 +25,17 @@ That is useful for safety, but it is not yet a complete retention policy. We wan
 
 ## Core Idea
 
-Treat retention as a policy on **artefact groups**, not individual tables.
+Treat retention as a policy on **artifact groups**, not individual tables.
 
 This keeps the model simple:
 
-- User-facing artefacts can have an optional TTL.
-- Processing data inherits the lifecycle of the artefact it belongs to.
+- User-facing artifacts can have an optional TTL.
+- Processing data inherits the lifecycle of the artifact it belongs to.
 - Analytics and durable user/account records are excluded unless we explicitly opt them in later.
 
 ## Proposed Retention Groups
 
-### 1. Batch Artefacts
+### 1. Batch Artifacts
 
 This group covers user-facing batch objects and the files attached to them.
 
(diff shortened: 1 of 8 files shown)


docs/data-retention-delete-policy.md
```diff
@@ -0,0 +1,218 @@
...
+This group covers internal processing state that exists only to execute or render batch work.
+
+Examples:
+
+- Requests
+- Request templates
+
+Policy:
+
+- Should not have an independent user-facing TTL.
```
pjb157: I think they are using facing as have customer info. Batch templates have the customer prompts and requests have the responses from the llms. Both are I think tied into batch arefacts

docs/data-retention-delete-policy.md
```diff
@@ -0,0 +1,218 @@
...
+Policy:
+
+- Out of scope for this proposal.
+- Continue to use existing deletion and scrubbing rules.
+
+## Deletion Model
+
+The policy should be easy to explain:
+
+1. A user-facing artefact may optionally be created with a TTL.
```
pjb157: There's also should be a system default clearly stated to the user

docs/data-retention-delete-policy.md
```diff
@@ -0,0 +1,218 @@
...
+It also aligns with the principle that user-visible artefacts should define the lifecycle, while internal processing data should follow behind them.
+
+## Open Questions
+
+These are the decisions worth discussing with others before implementation:
+
+1. Should batch/file TTL start from creation time or from terminal state?
+
+Recommended direction:
+Use terminal state where possible, so a long-running artefact does not expire while still active.
```
pjb157: agree

docs/data-retention-delete-policy.md
```diff
@@ -0,0 +1,218 @@
...
+
+2. Should there be a global default TTL for batch artefacts, or should TTL be opt-in only?
```
pjb157: Should be a global which is config driven and null in main

docs/data-retention-delete-policy.md
```diff
@@ -0,0 +1,218 @@
...
+
+Recommended direction:
+Start opt-in only.
+
+3. Should input files and output/error files share one policy, or should outputs have a different default?
```
pjb157, line 210: One policy I agree with recommendation

PR #1009, opened 2026-04-22
feat: upgrade fusillade 16.3.0 and switch async view to service_tier=flex

pjb157:
## Summary
- Upgrades fusillade from 16.2.0 to 16.3.0 (adds `service_tier` column with partial indexes for flex-tier queries)
- Switches the async requests view from filtering by `completion_window=1h` to `service_tier=flex`
- Removes `completion_window` column from the batch table UI
- Adds `service_tier` query parameter to `GET /admin/api/v1/batches/requests`
- Drops config dependency for async completion window in the dashboard

Batch creation still uses `completion_window` internally (fusillade API requirement). The "Batch only" toggle on the batches view still excludes 1h batches via `exclude_completion_window`.

## Test plan
- [x] `just lint rust` passes
- [x] `just test rust` passes (1186 tests, 0 failures)
- [x] `just lint ts` passes
- [x] `just test ts` passes (498 tests, 0 failures)
- [ ] CI passes

### DIFF
--- a/dashboard/src/components/features/batches/BatchesTable/columns.tsx
+++ b/dashboard/src/components/features/batches/BatchesTable/columns.tsx
@@ -33,8 +33,6 @@ interface ColumnActions {
   showUserColumn?: boolean;
   /** Show the Type column (hidden when "Batch only" filter is on) */
   showTypeColumn?: boolean;
-  /** Completion window that identifies async batches (default: "1h") */
-  asyncCompletionWindow?: string;
 }
 
 const getStatusColor = (status: BatchStatus) => {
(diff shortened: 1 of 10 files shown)


dashboard/src/components/features/batches/BatchesTable/columns.tsx
```diff
@@ -119,7 +117,7 @@ export const createBatchColumns = (
           header: "Type",
           cell: ({ row }: { row: { original: Batch } }) => {
             const batch = row.original;
-            const isAsync = batch.completion_window === (actions.asyncCompletionWindow ?? "1h");
```
pjb157, line 122: We allow platform managers to set any completion window so we should revert this

pjb157: Good catch — reverted in 053cf1ab. The type column is back to using the config-driven `asyncCompletionWindow` for detection. Only the `completion_window` display column was removed.

PR #1048, opened 2026-05-06
fix: Do not record requests as successful in outlet/http-analytics if they actually had errors mid-sse-stream

JoshC8C7:
## Summary

When an upstream inference engine crashes mid-stream, OpenAI-compatible providers (Dynamo, vLLM, etc.) send 200 OK headers and then emit a `data: {"error": {...}}` SSE frame in place of the terminal usage chunk. The chunk parser was silently dropping these frames. They failed to deserialize as `CreateChatCompletionStreamResponse` and got filtered out in `process_sse_chunks`, so analytics persisted the row as `status_code=200, total_tokens=0`, indistinguishable from a successful zero-token response.

Two visible effects:

- `http_analytics` rows from crashed-mid-stream attempts were counted as successful by every dashboard keyed on `status_code BETWEEN 200 AND 299` (per-batch `total_requests`, `user_model_usage` rollup, `get_user_model_breakdown_for_range`, etc.) and by the Prometheus `gen_ai_request_*` recorder, which gates `error_type` on `status_code >= 400`.
- The existing `CRITICAL: successful generative response recorded zero tokens` log fired on every fusillade retry, and fusillade can retry many times when the engine flaps.

### Changes

- `request_logging/models.rs`: add `StreamErrorChunk` and an `Error` variant on `ChatCompletionChunk` and `CompletionChunk` so the SSE parser retains the error frame instead of dropping it.
- `request_logging/serializers.rs`: in `UsageMetrics::extract`, when the upstream HTTP status is below 400 but the parsed body contains an `Error` chunk, override `status_code` to 502.
- `request_logging/analytics_handler.rs`: change the existing CRITICAL gate from `response_data.status.is_success()` to `(200..300).contains(&metrics.status_code)` so it no longer fires for rows we have already reclassified. The original safety net (genuine 200 with missing usage chunk) remains intact.

Credit deduction was already gated on `total_cost <= 0`, so zero-token rows continue to skip the credits path. No billing impact.


sejori (commented):
Makes sense to me - add fix: to the title tho

PR #1073, opened 2026-05-14
fix: filter responses and usage to org only for PMs

sejori:
aligns the filtering of responses and usage for PMs with that of batches. I.e. only show org data when in org context even if user has read_all perms.


sejori:
Consolidated into #1086. That PR brings in the Responses branch-order fix verbatim, both PM-in-org regression tests, and supersedes the Usage scoping change with a richer version that also adds an authorised `user_id` drill-down for org admins. Closing this so the staging deploy doesn't end up applying the same fix twice.

PR #1091, opened 2026-05-22
feat: add configurable browser security response headers

pjb157:
## Summary

Adds standard browser security response headers to every HTTP response, emitted directly by the server so they are present regardless of what sits in front of it (reverse proxy, ingress, or nothing). Defence in depth: a front layer that's misconfigured, replaced, or bypassed no longer means responses ship with no protections.

Controlled by a new `auth.security.headers` config section. Safe hardening headers are **on by default**:

- `X-Content-Type-Options: nosniff`
- `X-Frame-Options: DENY`
- `Referrer-Policy: strict-origin-when-cross-origin`
- `Permissions-Policy:` a locked-down default

`Content-Security-Policy`, its report-only variant, and `Strict-Transport-Security` are **opt-in** (empty by default): a CSP that doesn't match the deployed frontend can break it, and HSTS is usually owned by whatever terminates TLS. Operators set these to values appropriate for their deployment.

## Behaviour notes

- Headers use **if-not-present** semantics, so a stricter per-route header (e.g. `Referrer-Policy: no-referrer` on sensitive auth responses) is preserved rather than overwritten.
- Invalid configured header values **fail fast at startup** rather than being silently dropped.
- Implemented as `tower_http::set_header::SetResponseHeaderLayer` layers applied after CORS (added the `set-header` feature to the existing `tower-http` dependency).

## Notes for operators enabling CSP

If you set a strict `script-src 'self'` policy, note that a Vite-built frontend emits an inline module-preload polyfill by default; either disable that polyfill in the frontend build or include the appropriate source. Rolling a CSP out via `content_security_policy_report_only` first is recommended.

## Test plan

- [x] `cargo check -p dwctl` — clean
- [x] Unit tests added for the header builder (defaults on, disabled emits nothing, opt-in headers, invalid value rejected)
- [ ] CI: `just ci rust` (full test suite runs against Postgres)
- [ ] Manual: confirm headers present on a dashboard response and that per-route `Referrer-Policy` is preserved


pjb157:
Thanks for the thorough review. Responding to each point:

**1. Layer ordering / "OPTIONS preflight receives security headers" (marked blocking) — keeping as-is, intentionally.**

The security-header layers are applied as *outer* layers (after CORS) so every real response carries them — that's the defense-in-depth goal, and it matches the earlier review feedback asking for these to apply broadly. CORS preflight (`OPTIONS`) responses do also pick up the four inert headers, but that's harmless:

- Browsers only evaluate `Access-Control-*` on a preflight response; `X-Frame-Options`, `X-Content-Type-Options`, `Referrer-Policy`, and `Permissions-Policy` have no effect on a preflight and are simply ignored.
- Adding a response header can't *cause* a proxy to strip headers — that'd be pre-existing proxy behavior independent of this change. There's no failure mode here.

Moving the layers *inside* CORS to exclude preflights would (a) reverse the "apply to every response" intent from the prior review round, and (b) trade a real benefit (headers on all content responses, including any generated by outer middleware) for excluding an inert header set from `OPTIONS`. I'd recommend against that. Happy to reconsider if there's a concrete client/proxy that breaks, but I don't think one exists.

**2. Integration test asserting headers on a live response (non-blocking) — agree, good follow-up.**

The header *builder* is unit-tested (defaults present, disabled emits nothing, opt-in headers, invalid value rejected). A `TestServer`-based assertion that the headers land on a real response is a sensible addition; I'd do it as a small follow-up so we don't re-cycle this green PR.

**3. Example CSP policies (non-blocking).** `config.yaml` already ships a commented `content_security_policy` example; I can expand it with embedded-dashboard vs. separate-origin guidance if useful.

**4. `names()` test helper `.unwrap()` → `.expect(...)` (nit).** Fair — happy to switch it for clearer failure output.

On the CI side: the `sqlx-cli` pin to `^0.8` fixed the repo-wide `backend-test` failure (sqlx-cli 0.9.0 stopped auto-loading the per-crate `.env`, so `cargo sqlx prepare --check` couldn't find `DATABASE_URL`). All checks are green now.

Can fold 3 & 4 into a quick follow-up commit if you'd like them in before merge — otherwise this is ready.

PR #1096, opened 2026-05-26
feat: proxy provider-display-config icons through the API

pjb157:
## Summary

Adds `GET /admin/api/v1/provider-display-configs/{provider_key}/icon` — a server-side proxy that fetches the operator-set icon URL and streams the bytes back to the browser **under the same origin as the dashboard**.

## Why

`provider_display_configs.icon` URLs are operator-pasted in the admin surface and span many external hosts:

```
registry.npmmirror.com     ai2, alibaba, deepseek, lovable, mistral, nvidia, openai
www.elaia.com              .txt
upload.wikimedia.org       google
cdn.simpleicons.org        kimi
cdn.prod.website-files.com lighton
avatars.githubusercontent.com  moonshot
raw.githubusercontent.com  zai
```

…and that list grows each time an admin adds a provider with a new logo URL. Two CSP-side options are bad:

- **Enumerate hosts in `img-src`** → treadmill, requires a CSP hotfix every time a new host appears.
- **`img-src https:`** → allows any HTTPS image origin; widens the policy for every image load, not just provider icons.

This PR fixes it at the source: icons are fetched server-side and returned same-origin, so the SPA's CSP can keep `img-src 'self' data: blob:` and operators can configure any icon URL they like.

## Behaviour

| Case | Response |
|---|---|
| Valid `https://` icon, upstream serves an image | `200` with upstream bytes + `Content-Type` + `Cache-Control: public, max-age=3600` |
| No config / no icon / non-https URL | `404` (relative paths and registry-key shortcuts are handled client-side and bypass the proxy) |
| Upstream times out, returns non-2xx, or serves non-image | `500` |

Upstream-fetch safety:
- `https://` only
- **5 s total timeout**
- **2 MiB body cap**, enforced while streaming (no unbounded buffering)
- **3-redirect limit**
- **Content-Type must start with `image/`** (guards against an upstream returning an HTML error page that the browser would try to render in ``)

Auth: `RequiresPermission<Models, ReadOwn>` — same gate as `GET /provider-display-configs/{provider_key}`. The endpoint never accepts a URL from the client, only the `provider_key`, so SSRF is bounded to what an admin has already configured.

## Follow-ups (separate PRs)
1. **app-doubleword-private**: render `<ProviderBadge>` icons via this endpoint instead of the raw operator-set URL.
2. **internal**: once #1 ships, tighten the CSP `img-src` back to `'self' data: blob:`.

## Test plan
- [x] `cargo check -p dwctl` clean
- [x] My added code is rustfmt-clean (pre-existing fmt drift in main is unrelated)
- [ ] CI: `just ci rust` (clippy, fmt, tests)
- [ ] Manual: hit `/admin/api/v1/provider-display-configs/openai/icon` post-deploy with a real session, confirm the PNG/SVG renders

### DIFF
--- a/dwctl/src/api/handlers/provider_display_configs.rs
+++ b/dwctl/src/api/handlers/provider_display_configs.rs
@@ -373,9 +373,16 @@ pub async fn get_provider_display_config_icon<P: PoolProvider>(
         }
     };
 
+    // Several icon hosts (Wikimedia in particular, per its User-Agent policy)
+    // return 403 for requests with no User-Agent, so identify ourselves.
     let client = reqwest::Client::builder()
         .timeout(ICON_FETCH_TIMEOUT)
         .redirect(reqwest::redirect::Policy::limited(3))
+        .user_agent(concat!(
+            "dwctl-icon-proxy/",
+            env!("CARGO_PKG_VERSION"),
+            " (+https://github.com/doublewordai/control-layer)",
+        ))
         .build()
         .map_err(|e| Error::Other(anyhow::anyhow!("build http client: {e}")))?;
 
--- a/dwctl/src/api/handlers/provider_display_configs.rs
+++ b/dwctl/src/api/handlers/provider_display_configs.rs
@@ -16,6 +16,8 @@ use axum::{
     response::Response,
 };
 use std::collections::{BTreeMap, HashMap};
+use std::net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr};
+use std::sync::{Arc, OnceLock};
 use std::time::Duration;
 use url::Url;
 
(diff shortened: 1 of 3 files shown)


pjb157:
Pushed `3fa3018` addressing the review.

## Blocking (both rounds)

- **SSRF protection** — installed a custom `reqwest::dns::Resolve` that drops any address [`is_globally_routable`](https://github.com/doublewordai/control-layer/blob/feat/provider-icon-proxy/dwctl/src/api/handlers/provider_display_configs.rs) rejects: RFC1918, loopback, link-local (incl. `169.254.169.254`), CGNAT (`100.64/10`, also Tailscale), IPv6 unique-local/link-local, documentation/reserved. Filtering at the DNS layer rather than parsing the URL means CNAMEs and multi-record responses are covered too. I went with `std`'s stable predicates (`is_private`, `is_link_local`, `is_loopback`, …) plus hand-rolled bit checks for the ranges whose helpers are still nightly-only (`is_shared`, `is_unique_local`, `is_unicast_link_local`) so we don't need to pull in a new dep just for this. The DNS-rebinding angle is moot because the resolver decision is the connect decision — reqwest doesn't get a chance to swap IPs between them.
- **Redirects** — switched `Policy::limited(3)` → `Policy::none()`. The SSRF resolver only sees the initial host, so chasing a 302 into the internal network would bypass it. Any redirect now hits the existing `!status.is_success()` branch and returns 500.

## Non-blocking (Copilot)

- **Client reuse** — moved the client into a `OnceLock<reqwest::Client>` (`icon_http_client()`), so the connection pool, TLS state, and SSRF resolver are reused across requests.
- **Trim icon value** before `Url::parse`.
- **Drop duplicate `tracing::warn!` calls** at both error sites (lines 390 and 418) — `Error::Other` is logged at `error` level by the central `IntoResponse` impl, so we were double-logging.

## Tests

Added 11 pure-Rust unit tests for the SSRF predicate covering all the relevant ranges, IPv4-mapped IPv6 (`::ffff:127.0.0.1` etc), CGNAT boundary cases, and a couple of known-public addresses. I deliberately stopped short of mocking an HTTPS upstream to exercise the full handler path — happy to do that as a follow-up if you want, but the predicate is where the security-critical decision actually happens.

## Non-blocking I'm skipping

- **`utoipa` path nit**: every other handler in this file uses the relative-path convention (`/provider-display-configs`, `/provider-display-configs/{provider_key}/icon`) rather than the full `/admin/api/v1/...` prefix. The OpenAPI doc renders the prefix from the router mount-point, so the spec is correct. Worth aligning project-wide, but not as a one-off here.
- **Chunk bombs / slowloris (round 2)**: the 5-second `timeout` is the total time including body read, so a slow drip just hits the timer. And with a 2 MiB cap, even worst-case ~256-byte chunks bound the per-request allocator churn at ~8k chunks. Happy to revisit if we ever see this in practice.

pjb157:
Pushed `d076673` — two fixes I caught while exercising the proxy against all 13 production icon URLs locally.

## IP-literal hosts bypass the DNS resolver

When the URL host is already an IP (`https://10.0.0.1/...`, `https://169.254.169.254/...`, `https://[::1]/...`), reqwest goes straight to connect without consulting `dns_resolver`, so `SsrfSafeResolver` never runs. Added `url_host_is_globally_routable(url)` — a synchronous check that runs `is_globally_routable` on any `url::Host::Ipv4` / `Ipv6` and defers to the resolver for domain hosts. Invoked from the handler before the request fires, and from the redirect policy on every hop.

## `Policy::none()` was too aggressive

Two providers (npmmirror among them) serve a 302 to a CDN, so disabling redirects entirely returned 500 for them. Switched to `Policy::custom` capped at 3 hops with three checks per attempt: total count, https-only scheme, and `url_host_is_globally_routable`. Domain-based redirect targets still go through `SsrfSafeResolver` for the actual connect, so the IP-literal check is the only synchronous gap that needed plugging.

## Tests

3 more unit tests for the URL helper covering: hostnames pass through, private IP literals blocked (incl. `[::1]`, `[fc00::1]`, `[::ffff:10.0.0.1]`), public IP literals allowed. 14 SSRF tests total.

## Verified locally

- All 13 prod provider icons return 200 image bytes through the proxy at `:3001` (and via vite at `:5173`).
- `https://10.0.0.1/x` → 404 (handler URL-check, pre-flight).
- `https://localhost/x` → 500 (resolver rejects post-DNS).

PR #1101, opened 2026-05-27
feat: replace underway for response creation to use channel system

JoshC8C7:
## Summary

Replaces the two underway response jobs (`create-response`, `complete-response`) with an in-process channel and a batched writer task, modelled on the existing `AnalyticsBatcher`. Affects realtime only.

## Why

Every non-background realtime request enqueued two underway jobs. Under surge the queue backed up because each job paid per-row commits on `underway.task` on top of its own DB roundtrips. Underway is built for durable retries, and the realtime path doesn't need any: a dropped HTTP connection cannot be retried, and the row in `requests` is essentially a log for the responses listing. The analytics path already solves the same shape for `http_analytics` via a channel + batcher; this is the parallel for `requests`.

## Where this fits

| Path | Mechanism after this PR |
|---|---|
| Flex (`service_tier=flex`) | inline `create_flex` + daemon owns the lifecycle (unchanged) |
| Realtime (`/v1/responses` default/priority) | inline `create_realtime` for `background=true` + channel/writer for completion |
| Batch API (`/v1/batches`) | inline `create_batch_record` + underway `create-batch` job to populate (unchanged) |

Underway stays in the codebase for the batch API, which genuinely needs durable retry for the bulk populate-from-templates step. This PR only removes underway from the realtime completion path.

## New realtime flows

The writer is a single tokio task draining a `tokio::sync::mpsc::channel`. `FusilladeOutletHandler` resolves `created_by` from the api_key (one indexed lookup against the dwctl pool) and sends a `RawCompletedRequest` into the channel. The writer accumulates up to `batch_size` records and flushes them in one fusillade transaction via the new `Storage::persist_completed_realtime_batch` call.

### background=false (dominant path)

```text
client          middleware           upstream         channel        writer        fusillade.requests
  |                 |                   |                |             |                  |
  |---POST--------->|                   |                |             |                  |
  |                 |--proxy----------->|                |             |                  |
  |                 |<------response----|                |             |                  |
  |<---response-----|                   |                |             |                  |
  |                 |                                                                     |
  |                 | (outlet fires; handler resolves created_by, sends)                  |
  |                 |--send(RawCompleted).await------>  |             |                  |
  |                                                     |--recv------>|                  |
  |                                                                   |--INSERT (state='completed')-->|
```

No DB writes on the hot path. One fusillade transaction per batch, after the fact. Replaces what used to be two underway jobs.

### background=true

```text
client          middleware           upstream         channel        writer        fusillade.requests
  |                 |                   |                |             |                  |
  |---POST--------->|                   |                |             |                  |
  | (background)    |                   |                |             |                  |
  |                 |---create_realtime inline (state='processing')---------------------->|
  |<---202 resp_<id>|                                                                     |
  |                 |--spawn{ proxy --->|                |             |                  |
  |                                     |<--response-----|             |                  |
  |                                   (outlet fires from the spawn)                       |
  |                                     |--send(RawCompleted).await->|                   |
  |                                                                   |--UPDATE to 'completed'------>|
  |                                                                                                  |
  |---GET /responses/{id} (poll)                                                                     |
  |                 |<---read row-----------------------------------------------------------------|
  |<---response-----|                                                                               |
```

Inline `create_realtime` stays so the client's first poll finds the row. The completion side replaces the underway complete-response job with a channel send.

## What's removed

- `dwctl/src/responses/jobs.rs` (deleted)
- `create_response_job` and `complete_response_job` on `TaskRunner` and `TaskState` plus the worker spawns in `TaskRunner::start`
- `create_response_job` field on `ResponsesMiddlewareState`
- `response_workers` config field (replaced by `response_writer_batch_size`, default 100)
- The create-response enqueue branch in `handle_realtime`'s non-background path

## What's added

- `dwctl/src/responses/writer.rs` with `RawCompletedRequest`, `RequestsWriterSender`, `RequestsWriter` (modelled on `AnalyticsBatcher::run`)
- `requests_writer_sender` field on `BackgroundServices` and a new parameter on `build_router`
- `response_writer_batch_size: usize` config field (default 100)
- Metrics: `dwctl_requests_writer_channel_depth`, `dwctl_requests_writer_sends_total{result}`, `dwctl_requests_writer_dropped_total{reason}`, `dwctl_requests_writer_flush_duration_seconds`, `dwctl_requests_writer_flush_size`, `dwctl_requests_writer_flush_errors_total`, `dwctl_requests_writer_records_total`, `dwctl_requests_writer_retries_total{outcome}`

## Trade-offs

1. **Records can be lost on dwctl crash.** Same trade as `http_analytics`. For non-background realtime the client lost the connection anyway. For background realtime a poll will 404, matching the pre-existing failure mode on the old underway path.
2. **No in-flight `processing` row for non-background realtime.** The client cannot poll while holding the connection. Only admin observers would have seen this row; the responses listing now shows it once it completes.
3. **Records without a resolvable api_key are dropped at the outlet handler.** Same behaviour as the underway era (the old create-response job returned early on missing/invalid api_key); the skip just moved from inside the worker to the outlet handler.

## Dependencies

Depends on fusillade introducing `Storage::persist_completed_realtime_batch` (separate PR: `fusillade/pr-persist-completed-realtime-batch.md`). A local `[patch.crates-io] fusillade = { path = "../fusillade" }` is in `application/Cargo.toml` for testing. **Remove the patch and bump fusillade in `dwctl/Cargo.toml` before merging.**

## Existing underway tasks at deploy time

The old `create-response` and `complete-response` queues stop having any workers after rollout. Pending rows for those queues in `underway.task` are orphaned (nothing reads them, nothing writes them). Functionally invisible; clean up after rollout with `DELETE FROM underway.task WHERE task_queue IN ('create-response', 'complete-response')` once the deploy is stable. Realtime responses whose underway tasks were still pending at deploy time lose their `requests` row only; `http_analytics` still captures them.


sejori (approved):
LGTM

PR #1125, opened 2026-06-04
fix(openapi): use Doubleword-served models in API doc examples

pjb157:
## What

Updates the model names used in the OpenAPI / utoipa **example values** so the rendered API docs reference models we actually serve instead of OpenAI's:

- **Responses API** (`ResponseRequest` / `ResponseObject` examples): `gpt-4o` → `deepseek-ai/DeepSeek-V4-Pro`
- **Embeddings API** (`EmbeddingRequest` / `EmbeddingResponse` examples): `Qwen/Qwen3-30B-A3B-FP8` → `Qwen/Qwen3-Embedding-8B`
- **Retrieve model** endpoint param description: `gpt-4`, `text-embedding-ada-002` → `deepseek-ai/DeepSeek-V4-Pro`, `Qwen/Qwen3-Embedding-8B`

## Why

Our API documentation should showcase the models we serve, not OpenAI model names. As a bonus, the embeddings endpoint examples were previously showing `Qwen/Qwen3-30B-A3B-FP8` — a chat model — as the embedding model; they now use a real embedding model (`Qwen/Qwen3-Embedding-8B`).

## Scope / reviewer notes

- **Documentation-only.** Every change is a `#[schema(example = ...)]` value or a `#[utoipa::path(...)]` param description. No runtime behavior changes, no schema/field changes.
- **Intentionally left unchanged:** chat-completion, model-listing, and batch/files cost examples still use `Qwen/Qwen3-30B-A3B-FP8` (a valid chat model, not an embedding reference). Test fixtures / sample data that reference `text-embedding-*` model names were left alone — they're test code, not docs (and `strict_mode.rs` ties the model name to onwards test setup).
- The only OpenAPI test (`test_openapi_specs_serialize`) just asserts the spec serializes and contains `"openapi"`; it does not assert on example values, so it's unaffected. `cargo fmt --check` passes.

### DIFF
--- a/dwctl/src/openapi/ai.rs
+++ b/dwctl/src/openapi/ai.rs
@@ -121,7 +121,7 @@ fn list_models() {}
     summary = "Retrieve model",
     description = "Retrieves information about a specific model.",
     params(
-        ("model" = String, Path, description = "The model ID (e.g., `gpt-4`, `text-embedding-ada-002`)")
+        ("model" = String, Path, description = "The model ID (e.g., `deepseek-ai/DeepSeek-V4-Pro`, `Qwen/Qwen3-Embedding-8B`)")
     ),
     responses(
         (status = 200, description = "Model details including ID, owner, and creation timestamp.", body = extra_types::ModelObject),
--- a/dwctl/src/openapi/extra_types.rs
+++ b/dwctl/src/openapi/extra_types.rs
@@ -308,12 +308,12 @@ pub struct Usage {
 /// Request body for embeddings.
 #[derive(Debug, Clone, Serialize, Deserialize, ToSchema)]
 #[schema(example = json!({
-    "model": "Qwen/Qwen3-30B-A3B-FP8",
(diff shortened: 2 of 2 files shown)


pjb157:
Re: the `doubleword-code` review flagging `deepseek-ai/DeepSeek-V4-Pro` as blocking ("model doesn't exist in configured models") — keeping it as-is, intentionally.

The OpenAPI `#[schema(example = ...)]` values here are **illustrative**, not tied to the dashboard's demo mock data (`models.json`). Two concrete points:

- The pre-existing chat-completion example `Qwen/Qwen3-30B-A3B-FP8` is **also** absent from `models.json`, so "present in the mock data" was never the bar for example values in these docs.
- `deepseek-ai/DeepSeek-V4-Pro` is a served model per docs.doubleword.ai — as verified in the **earlier** automated review on this same PR, which noted the demo mock data is intentionally incomplete vs production.

This PR is documentation-only (no runtime/schema change), and the example IDs are illustrative. No change needed.

PR #1150, opened 2026-06-16
fix(image-normalizer): don't re-sign URLs already in our own store

pjb157:
## Problem

When image-input normalization is enabled, the realtime `image_normalizer_middleware` runs in `Mode::All`, which matches `http(s)` image URLs.

Requests dispatched through the batch path have already had their `dw-img://` tokens resolved to signed object-store URLs by the dispatch-time JIT signer, which uses the **dispatch TTL** (derived from the processing timeout plus headroom, so a signed URL always outlives a dispatch attempt).

Because those already-signed URLs are plain `https://…` URLs, the realtime middleware treated them as fresh external inputs: it **re-ingested and re-signed them with the much shorter realtime TTL**. The net effect was that batch-dispatched image URLs reached the upstream model carrying the realtime TTL instead of the intended dispatch TTL.

If the upstream worker doesn't fetch the image promptly — for example when it is backlogged — the short-lived URL can expire before the fetch, and the object store responds `403 Forbidden`. This surfaces as an image-loading failure in the worker and a failed request.

## Fix

Teach the store to recognise **its own** signed URLs and have the realtime middleware pass them through unchanged instead of re-ingesting and re-signing:

- `ImageStore::owns_url(&str) -> bool` (default `false`), implemented for the S3-compatible, GCS, and in-memory backends.
- Exposed via `ImageNormalizer::owns_url`, delegating to the store.
- The middleware's substitute closure now skips (forwards unchanged) any non-`data:` URL for which `owns_url` is true.

This preserves the upstream (dispatch) TTL and avoids a redundant re-fetch of an image we already host. Genuine external / `data:` image inputs are unaffected and still normalized as before.

`owns_url` is **host- and bucket-bound** (S3/GCS: endpoint or host plus the `/{bucket}/images/` key prefix; in-memory: the configured base URL), so an external URL that merely mimics the content-addressed key path is **not** mistaken for one of ours.

## Tests

- `store.rs`: `owns_url` for the S3-compatible backend (including a spoofed-external-host negative and trailing-slash endpoint handling), GCS, and in-memory.
- `image_normalizer_middleware.rs`: regression test asserting a URL already in the store is forwarded **unchanged** (no re-sign / TTL change).

> The wider `dwctl` crate requires a database for its `sqlx` query macros to compile; the change here is confined to the database-free `image_normalizer` module and middleware, so the added tests are exercised by CI.

## Possible follow-ups
- Consider deriving the batch dispatch URL TTL from the batch completion window rather than the per-attempt processing timeout, so very long-queued batches cannot expire their URLs even when correctly signed once.
- A worker that holds a request far longer than the per-attempt timeout (e.g. while recovering from a restart) can still outlive even a correctly-sized URL; capping per-request image count/size and ensuring adequate worker memory reduces that risk.

### DIFF
--- a/dwctl/src/image_normalizer/mod.rs
+++ b/dwctl/src/image_normalizer/mod.rs
@@ -136,6 +136,15 @@ pub trait ImageNormalizer: Send + Sync {
     /// Read bytes for `token` directly. Used by the dashboard image
     /// endpoint (after authorisation).
     async fn read(&self, token: ImageToken) -> Result<(String, Bytes), NormalizeError>;
+
+    /// True if `url` already points at an object in our own store (a URL we
+    /// previously signed). Callers use this to avoid re-ingesting/re-signing
+    /// an already-normalised URL — which would waste a re-fetch and clobber a
+    /// longer upstream TTL (e.g. the batch dispatch TTL) with a shorter one.
+    /// Default `false` preserves prior always-ingest behaviour.
+    fn owns_url(&self, _url: &str) -> bool {
+        false
+    }
 }
 
 /// No-op normaliser used when `config.enabled = false`. Surfaces an error
@@ -221,6 +230,10 @@ impl<S: ImageStore + 'static> ImageNormalizer for DefaultImageNormalizer<S> {
 
     async fn read(&self, token: ImageToken) -> Result<(String, Bytes), NormalizeError> {
         Ok(self.store.read(token).await?)
+    }
+
+    fn owns_url(&self, url: &str) -> bool {
+        self.store.owns_url(url)
     }
 }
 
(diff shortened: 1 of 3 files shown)


pjb157:
Thanks @copilot — good catches, all three addressed in 911a627f.

The `owns_url` checks no longer do raw `starts_with`/`contains` on the URL string. All three backends now parse the URL (`url` crate) and:
- match the **host exactly** — so `storage.googleapis.com.evil.com` / `…r2.cloudflarestorage.com.evil.com` suffix-spoofs are rejected;
- compare scheme (and port where relevant);
- look for the `images/` content-addressed prefix on the **parsed path only**, never the query/fragment;
- require path-prefix matches to land on a `/` boundary, so `/dw-img2` and `/dw-img.evil` no longer match `/dw-img`.

Also fixed the false-negative you flagged on `S3CompatStore`: it now recognises **virtual-hosted** presigned URLs (`{bucket}.{endpoint-host}/images/…`) as well as path-style, so it works when `force_path_style = false`.

Tests extended with suffix-spoof, query-injection, sibling-prefix, wrong-scheme, and virtual-hosted cases.

PR #1155, opened 2026-06-17
feat(cors): optional any-origin CORS without credentials

pjb157:
## Summary

Adds an optional CORS setting, `auth.security.cors.allow_any_origin_without_credentials` (default `false`). It is purely additive — with the flag unset, CORS behavior is unchanged.

When enabled:
- Any request origin is reflected in `Access-Control-Allow-Origin`, with no `Access-Control-Allow-Credentials`.
- Origins in the configured `allowed_origins` allowlist continue to receive credentialed CORS, exactly as before.

This makes the API's CORS origin/credential handling configurable. Per the Fetch spec a wildcard origin and credentials are mutually exclusive, so credentials remain scoped to the explicit allowlist.

## Changes
- New `allow_any_origin_without_credentials` field on `CorsConfig` (default `false`).
- `create_cors_layer` / `apply_cors` apply a reflective, credential-less CORS layer when enabled, plus a small middleware that re-grants credentials to allowlisted origins.

## Testing
- Unit tests across the origin × credentials matrix (allowlisted, arbitrary, look-alike, no `Origin`) and config parsing/default.
- `cargo fmt` and `cargo clippy` clean.


pjb157:
Thanks for the reviews — addressed in c3b53603:

- Renamed `credentialed_origins` → `origins_to_grant_credentials`, and documented why wildcard entries are intentionally skipped (a wildcard can't carry credentials per the Fetch spec).
- Added a startup log when `allow_any_origin_without_credentials` is enabled, so the active mode is visible operationally.
- Added tests for actual (non-preflight) `GET` requests from both third-party and allowlisted origins, complementing the existing preflight coverage.
- Re-ran `cargo fmt` — this also resolves the `backend-test` formatting failure.

Left as-is: the `.cloned()` in the credentials middleware (negligible cost, keeps it readable). The config field already documents the behavior, and the startup log now covers operator visibility of the mode.

PR #1168, opened 2026-06-22
feat(responses): fetch and inline input_file file_url for document inputs

JoshC8C7:
# feat(responses): fetch and inline input_file file_url for document inputs

## Summary

This is the dwctl half of COR-427 (PDF / `input_file` support on
`/ai/v1/responses`). The onwards half makes the gateway able to carry and map
`input_file` content; this half makes a `file_url` actually work end to end in
strict mode.

A Chat Completions `file` content part can only carry inline bytes
(`file_data`) or a provider-owned `file_id` - it has no URL form (unlike
`image_url`). So an `input_file.file_url` cannot survive the
Responses-to-Chat-Completions conversion as a URL: the bytes have to be fetched
and inlined first. dwctl is the only layer that can do that fetch (onwards is a
pure transform with no outbound I/O).

A new middleware rewrites each `input_file` content part in a `/responses`
request body before it reaches onwards:

- `file_url` -> fetch through the hardened fetcher (the same SSRF-protected
  fetcher used for image URLs: DNS pinning, IP deny-list, redirect
  re-validation, MIME / size caps) and inline the bytes as a base64 `file_data`
  data URI, dropping `file_url`.
- `file_data` -> left untouched; onwards maps it to a Chat Completions `file`
  part.
- `file_id` (with no `file_data`) -> rejected with a clear 400. dwctl customers
  never upload to the upstream provider, so a bare `file_id` always refers to
  dwctl's own file store, which is not yet wired to resolve Responses inputs.
  Returning a documented error beats forwarding an id the provider can never
  resolve.

## Changes

- New `crate::file_input` module: `FileInputConfig` (enabled flag + a
  document-MIME fetcher policy, default `application/pdf`) and
  `normalize_input_files`, which walks `input[*].content[*]` and rewrites
  `input_file` parts. The fetch is injected as a callback (mirroring
  `image_normalizer::walker::substitute_with`) so the rewrite logic is testable
  without a live fetch.
- New `inference::file_input_middleware`: reads the body once, runs
  `normalize_input_files` with the real hardened fetcher, restores / reserialises
  the body, and maps failures to file-specific 4xx/5xx codes
  (`input_file_rejected`, `input_file_url_unfetchable`, ...). A raw-body
  substring fast-path skips the JSON parse entirely unless the body actually
  mentions `input_file`, and a body with nothing to rewrite is restored
  byte-for-byte (no reserialise, no Content-Length change).
- Reuses `image_normalizer::fetcher::ImageFetcher` and `NormalizeError`
  verbatim - only the MIME allow-list differs (documents instead of images).
- `Config` gains a `file_input` section; the middleware is layered onto the
  onwards router next to the image normaliser.

## Behaviour by config

- `file_input.enabled = false` (default): the middleware is a pass-through. An
  `input_file.file_url` flows to onwards untouched and onwards returns a clear
  strict-mode error rather than silently dropping it. Outbound fetch of
  user-supplied URLs is opt-in, mirroring the image normaliser.
- `file_input.enabled = true`: `file_url` is fetched and inlined, so a PDF by
  URL works end to end through strict mode.

## Tests

`crate::file_input` (rewrite logic): fetch+inline `file_url` to base64
`file_data`, `file_data` passthrough (asserts no fetch), bare `file_id`
rejected, no-`input_file` no-op, and a real-fetcher SSRF rejection
(link-local). `inference::file_input_middleware` (HTTP layer): `file_id` 400,
link-local `file_url` 400, inline `file_data` passthrough, no-`input_file`
verbatim passthrough, disabled passthrough, and non-`/responses` paths ignored.
`cargo test -p dwctl --lib file_input` passes (11 tests).

Note on test strategy: the SSRF guard hard-denies loopback, so a local mock
server cannot be fetched. The success path is covered by injecting a stub fetch
(the production fetcher is exercised separately against a link-local address to
prove the guard fires).

## Dependency / deploy ordering

This change has no compile-time dependency on the onwards half: dwctl rewrites
the JSON body (producing `input_file.file_data`) and forwards it, so it builds
against the current released onwards. The runtime dependency is that onwards
must understand `input_file.file_data` for the inlined document to reach the
model - that is the onwards half of COR-427.

So this is safe to merge independently: the feature is off by default
(`file_input.enabled = false`), and it is dormant until both halves are
deployed and the flag is enabled. Do not enable `file_input` in an environment
until the onwards `input_file` change is live there.

(Locally this branch is tested end to end against the onwards branch via a
`[patch.crates-io]` override in the workspace `Cargo.toml`. That override is
intentionally NOT part of this PR.)

## Follow-up

dwctl-owned `file_id` resolution: intercept a dwctl-issued `file_id`, load the
bytes from dwctl's file store, and inline them as `file_data` (the same shape
this middleware already produces for `file_url`). That turns the `file_id` 400
into a working path and is the remaining piece of COR-427.

### DIFF
--- a/dwctl/src/config.rs
+++ b/dwctl/src/config.rs
@@ -200,6 +200,13 @@ pub struct Config {
     /// shape and defaults.
     #[serde(default)]
     pub image_normalizer: crate::image_normalizer::ImageNormalizerConfig,
+    /// `input_file` (PDF / document) handling for `/v1/responses`. When
+    /// enabled, an `input_file.file_url` is fetched through the hardened
+    /// fetcher and inlined as base64 `file_data` before forwarding, so it
+    /// survives the strict Responses-to-Chat-Completions conversion. See
+    /// `crate::file_input` for the full shape and defaults.
+    #[serde(default)]
+    pub file_input: crate::file_input::FileInputConfig,
     /// OpenAPI spec exposure controls. Defaults disable the Admin spec
     /// (which describes internal management endpoints) and enable the
     /// AI spec (which mirrors the publicly-documented OpenAI surface).
@@ -1968,6 +1975,7 @@ impl Default for Config {
             connections: ConnectionsConfig::default(),
             responses: ResponsesConfig::default(),
             image_normalizer: crate::image_normalizer::ImageNormalizerConfig::default(),
+            file_input: crate::file_input::FileInputConfig::default(),
             openapi: OpenApiConfig::default(),
         }
     }
--- a/dwctl/src/db/handlers/api_keys.rs
+++ b/dwctl/src/db/handlers/api_keys.rs
@@ -2068,6 +2068,7 @@ mod tests {
             connections: Default::default(),
             responses: Default::default(),
             image_normalizer: Default::default(),
+            file_input: Default::default(),
             openapi: Default::default(),
         };
         crate::seed_database(&config.model_sources, &pool).await.unwrap();
--- a/dwctl/src/file_input/mod.rs
+++ b/dwctl/src/file_input/mod.rs
@@ -1,0 +1,231 @@
+//! Normalisation of Responses `input_file` content (PDF / document inputs).
+//!
+//! The OpenAI-compatible Chat Completions `file` content part can only carry
+//! inline bytes (`file_data`) or a provider-owned `file_id` - it has no URL
+//! form (unlike `image_url`). So an `input_file` submitted to `/v1/responses`
+//! with a `file_url` cannot survive the Responses-to-Chat-Completions
+//! conversion as a URL: the bytes have to be fetched and inlined first.
+//!
+//! This module rewrites each `input_file` content part in a Responses request
+//! body before it reaches onwards:
+//!
+//! - `file_url` -> fetch through the hardened [`ImageFetcher`] (the same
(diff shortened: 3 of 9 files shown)


JoshC8C7:
Thanks for the thorough review - addressed the coverage gap and the nits.

## Coverage (the main one)
`input_file` now works on **all** paths, not just single-step:
- `transition.rs`: maps `input_file{file_data|file_id}` to a Chat Completions `{"type":"file","file":{...}}` part instead of dropping it.
- `inference` middleware: inlines `input_file.file_url` (via the hardened SSRF fetcher) on the **warm-path** (multi-step / tools) and **flex** branches, before the request is dispatched / persisted. The flex job is persisted with `file_data` already inlined, so the daemon needs no extra fetching. Realtime continues to use the `file_input` layer.

So multi-step and flex now carry the document end to end.

## Nits
- `file_input/mod.rs`: `.expect()` -> returns `NormalizeError::BadInput` on a malformed part.
- Documented the PDF-only MIME default (operators can widen `allowed_mime`) and the ~33% base64 body inflation (keep `max_bytes` aligned with upstream limits).
- Softened the fast-path comment: it is best-effort, not a correctness boundary. A JSON-escaped discriminator skips normalisation and the `file_url` flows to onwards, which returns a clear error (never a silent drop).
- memchr micro-opt and the config doc-wording: left as-is (no functional change).

PR #1208, opened 2026-06-30
fix(perf): defer cache classify-await into the SSE stream so it never holds the first token

hachall:
## What & why

For streaming chat-completions, the cache layer was **joining the classify task under the
5s deadline *before* returning the response** — so if classify stalled (tokenizer/index
hiccup), the client's **first token was held up to 5s**, even though the cache stats are
only spliced into the **terminal usage frame**.

This defers the classify-await into the SSE stream: the response returns immediately (first
token flows untouched), and classify is resolved lazily at the terminal usage frame. Since
classify is `spawn`ed, it runs concurrently the whole time the model generates, so by the
last frame it's almost always already done — the await is instant in the normal case and
bounded by the deadline only on an actual outage, where it can delay at most the **final**
frame. (This is §19's top deferred follow-up.)

## How

Post-response handling now splits on transport:

- **Non-streaming** — unchanged. The body is fully buffered when `next.run` returns, so
  classify already finished during generation; we join it inline (instant) and edit the JSON.
- **Streaming** — `defer_classify_into_stream` wraps the SSE body in an `async_stream` that
  streams frames straight through, awaits the classify handle at the terminal usage frame
  (bounded by the deadline), injects the stats there, and commits the index write on a
  billing-success completion.

## Fail-safe

Every failure path degrades to **no caching, request unharmed** — never a blocked, failed,
or hung request:

- classify deadline / error / panic → terminal frame emitted unedited, no commit;
- mid-stream error frame or no usage frame → veto, no commit;
- client disconnect → the stream is dropped before the commit code runs → no commit;
- the deadline `abort()`s the spawned task; the request body and all earlier frames are
  never touched.

The resolved `ClassifyOutcome` at the last frame is identical to what the old inline join
produced, so injection (including the uniform-zeros for enabled models) is byte-for-byte
unchanged — just computed later.

## One conscious tradeoff

On an **early client disconnect**, the dropped async-stream correctly does **not** commit,
but it also no longer records the `commit_vetoed{reason="stream_aborted"}` *metric* for that
case (the old `Drop`-based `VerdictStream` did). Correctness is preserved; only that one
observability counter under-counts on disconnects. Can add a drop-guard if we want it back.

## Changes

- `layer.rs`: split the post-response flow; added `join_classify`, `outcome_label`,
  `is_streaming`, and `defer_classify_into_stream`.
- `inject.rs`: removed `VerdictStream` + `CommitGate` + the SSE branch; renamed the entry to
  `inject_into_response_nonstreaming` (returns `(Response, bool)`); exposed `scan_inject_sse`
  as `pub(crate)`.
- `mod.rs`: trimmed re-exports.

## Testing

- New layer-level streaming end-to-end test: SSE upstream → first stream creates + commits,
  second stream reads the cached prefix — both injected into the terminal usage frame, with
  deltas and `[DONE]` preserved.
- Reworked inject tests: direct injection-primitive coverage + non-streaming success/veto.
- Full `prompt_cache` suite green (59 tests), clippy + fmt clean. No SQL changes.


hachall:
## Disconnect handling

On an early client disconnect the dropped async-stream correctly does **not** commit, and an
`AbortOnDrop` guard cancels the in-flight classify task (no orphaned/detached task). The only
gap is the `commit_vetoed{reason="stream_aborted"}` *metric*, which isn't recorded for that
case — correctness is preserved; only that one observability counter under-counts on disconnects.

PR #1210, opened 2026-07-01
fix: optimize onwards config balance queries

pjb157:
## Summary
- materialize user balance CTEs used while building routing configuration
- exclude deleted users from balance eligibility CTEs
- refresh SQLx metadata and add a regression check for the query shape

## Testing
- cargo fmt --check
- cargo check -p dwctl --no-default-features
- cargo test -p dwctl --no-default-features sync::onwards_config::tests::test_user_balance_ctes_are_materialized_and_skip_deleted_users -- --nocapture
- cargo sqlx prepare --check --workspace -- --no-default-features


dwctl/src/sync/onwards_config/mod.rs
hachall, line 603: i'm not sure how much materialization is going to make a difference, as the user_balance CTE is only used once in the wider query. could even slightly slow it down if it tries to write it to disk before starting the next part of the query  

"Overhead When Not Needed: If a CTE or subquery is only used once and is simple, materializing it can waste memory or disk space. Inlining is often faster in these cases, as it skips the storage step." [link](https://medium.com/@pritammanijoy2001/materialized-made-simple-your-guide-to-smarter-sql-with-ctes-and-views-2b5fcc427c22)

hachall (approved):

dwctl/src/sync/onwards_config/mod.rs
hachall: unless it recreates user_balances every time it attempts this clause, then definitely would be useful

hachall: ok definitely materialize - I ran explain analyze with as MATERIALIZE and NOT MATERIALIZE, and former is 4x faster

pjb157: Haha trust the vibes

PR #1214, opened 2026-07-01
feat: account-wide zero data retention toggle for users and orgs

sejori:
## Summary

Adds an account-wide **Zero Data Retention (ZDR)** toggle for users and organizations, and plumbs the flag from the `users` table onto the API keys that the `onwards` sync consumes.

This is the **control-layer side only**. onwards receives the flag but does **not** act on it yet — that's a deliberate follow-up. The setting is stored on the account and surfaced to onwards as a per-key `zdr` label.

### Design

- **One column, both entities.** Organizations are rows in the same `users` table (`user_type = 'organization'`), so a single `users.zero_data_retention` column covers users *and* orgs. This mirrors the existing `verified` flag (migration 099).
- **Per-owner cascade for free.** The onwards sync already does `JOIN users u ON u.id = ak.user_id`. Org-context keys carry `user_id = org.id`, so enabling ZDR on an org enables it for every key created within the org; a member's personal keys are governed by their own flag. No extra logic.
- **Zero onwards changes.** onwards' `KeyDefinition` already carries a `labels` map (currently `{"purpose": …}`). The sync now also emits `{"zdr": "true"|"false"}`. onwards stores it and does nothing with it yet — exactly the intended scope. The label is **always emitted** so future routing rules can match either value; it is a no-op when `false`.
- **Admin-only.** Only callers with `UpdateAll` (on users / organizations) may set the flag — an account holder cannot flip ZDR on themselves. Enforced in both the user and organization PATCH handlers.

### Changes

**Migration**
- `108_add_zero_data_retention.sql` — adds `users.zero_data_retention BOOLEAN NOT NULL DEFAULT false` + a column-scoped `AFTER UPDATE` trigger that fires `notify_config_change()` so toggles rebuild the onwards cache immediately (mirrors `users_verified_notify`).

**Backend (dwctl)**
- Threads the field through the `User` DB entity, `UserDBResponse`, `UserResponse`, `UserUpdate(DBRequest)`, and the organization equivalents.
- `update_user` / `update_organization` reject the field for non-admins (like role escalation).
- onwards sync (`sync/onwards_config`): both key-selection queries now project `u.zero_data_retention`; `OnwardsApiKey` carries it; both label builders emit the `zdr` label.

**Dashboard**
- ZDR toggle in **Edit User** and **Edit Organization** modals (admin-gated; the org toggle only renders on the User Access page via `canEditZdr`).
- `zero_data_retention` added to `User`, `UserUpdateRequest`, `OrganizationUpdateRequest`.

### Tests

- `test_admin_can_toggle_zero_data_retention` — admin can enable/disable; round-trips.
- `test_standard_user_cannot_set_own_zero_data_retention` — 403, and nothing persists.
- `test_cache_shape_zero_data_retention_label_reflects_owner` — onwards sync emits `zdr=true` for a ZDR owner's key and `zdr=false` for a non-ZDR owner's key.
- Frontend: existing user/org/access component suites pass (53 tests); new org-modal path covered.

### Notes / follow-ups

- onwards does not yet consume the `zdr` label — separate PR.
- Draft: opening for review of the plumbing approach before wiring onwards behavior.


sejori:
Screenshot 2026-07-02 at 09 36 32

(Same toggle available on org edit as well)

pjb157:
Can we add `zero_data_retention: bool` to the org membership summary returned from the API? Just building something in the private UI to show users if they have it on or off?

PR #1217, opened 2026-07-02
feat: surface zero data retention to org members (summary + org/profile pages)

sejori:
## Summary

Follow-up to #1214. Makes zero data retention (ZDR) status **visible to org members** (read-only), and exposes the flag on the org membership summary that the private UI needs.

### API

- `OrganizationSummary` (the per-org entry in a user's `organizations` list, from `GET /users/{id}/organizations` and `?include=organizations`) now carries the org's `zero_data_retention`. Both build sites already map from the org's `UserDBResponse`, which selects the column — **no migration, SQL, or `.sqlx` change**.

### Dashboard (member-facing, read-only)

- **Organization page** (`MyOrganization`): a status badge in the header showing whether ZDR is enabled for the org (reads the org's `zero_data_retention`).
- **Profile page** (`Profile`): a "Data: Zero retention" row in Account Details showing the user's own account ZDR status, with an explanatory hover.

Both consume data the API already returns (the full `Organization` / `User` objects carry `zero_data_retention` from #1214); no new endpoints. The admin-only toggle to *change* ZDR remains from #1214 — these views are read-only.

### Tests

- Backend: `test_user_organizations_summary_reflects_org_zero_data_retention` (summary reports `false` by default, `true` after an admin enables ZDR on the org).
- Frontend: new cases assert the org-page badge shows "Zero data retention on" and the profile row shows "Zero retention" when enabled; `tsc` + eslint clean; affected suites pass (28 tests).


sejori:
Screenshot 2026-07-02 at 16 31 22
Screenshot 2026-07-02 at 16 44 58

(false states are not displayed for now)

PR #1225, opened 2026-07-06
feat: set realtime priority value to 0

sejori:
- **Created** the inline nvext mutation as a pure, testable helper `attach_realtime_priority(&mut serde_json::Value)` — mirroring how `resolve_service_tier` / `warm_path_branch` are structured for testability. The Realtime arm calls it before sending on to onwards.
- **Added 6 unit tests** asserting realtime traffic always ends up with `nvext.agent_hints.priority == 0`:
  - bare body → priority 0
  - representative realtime shapes (responses / chat completions / embeddings, with tools, with pre-existing unrelated `nvext`, empty body) all get priority 0
  - pre-existing sibling `nvext` fields preserved; `agent_hints` replaced with exactly the hint
  - non-object `nvext` coerced without panic, still priority 0
  - non-object body is a safe no-op

### DIFF
--- a/dwctl/src/inference/middleware.rs
+++ b/dwctl/src/inference/middleware.rs
@@ -1495,6 +1495,17 @@ mod tests {
     }
 
     #[test]
+    fn attach_realtime_priority_replaces_non_object_agent_hints() {
+        let mut body = serde_json::json!({
+            "model": "m",
+            "nvext": { "guided_json": { "x": 1 }, "agent_hints": "surprise" },
+        });
+        attach_realtime_priority(&mut body);
+        assert_eq!(body["nvext"]["guided_json"], serde_json::json!({ "x": 1 }));
+        assert_eq!(body["nvext"]["agent_hints"], serde_json::json!({ "priority": 0 }));
+    }
+
+    #[test]
     fn attach_realtime_priority_coerces_non_object_nvext() {
         // A hostile/odd client sending a non-object nvext must not panic
         // and must still end up with priority 0.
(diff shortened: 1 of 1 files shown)


sejori:
Rebased onto `main` (resolved the conflict with the new realtime ZDR handling) and addressed the review feedback:

- **Copilot — hint not sent downstream (`middleware.rs:321`):** fixed. `attach_realtime_priority` mutated `request_value`, but `handle_realtime` rebuilt the forwarded request from the original `body_bytes`, so `nvext.agent_hints.priority=0` only reached the stored tracking row and never onwards/the scheduler. Now the Realtime arm rebuilds `body_bytes` from the mutated `request_value` before dispatch, and `handle_realtime` sets `Content-Length` to the rewritten body length so the backend doesn't truncate/reject.
- **Copilot — test count:** the branch now has 6 `attach_realtime_priority` tests (incl. `replaces_non_object_agent_hints`, the non-object `agent_hints` case), matching the description. All 6 pass; `cargo clippy -p dwctl --lib` is clean.

Context: this targets realtime requests being starved behind batch work on Dynamo backends (batches carry `-batch_expires_at` priorities; realtime was unprioritised), which manifested as realtime hangs on `gemma-4-31B-it` / `GLM-5.2-FP8`.

Note/possible follow-up: the priority hint is attached on the `handle_realtime` fall-through path; tool-using `/v1/responses` realtime requests dispatched via the multi-step warm path don't yet get it — worth a follow-up if those also need to preempt batch.

PR #1247, opened 2026-07-09
fix: dont flatten asst and tool result blocks in anthropic adapter

hachall:


hachall:
### Why: the adapter drops the SDK's cache breakpoints

`convert_assistant` / `convert_user` flatten assistant and `tool_result` blocks to plain **strings**. That discards any `cache_control` on them — and the Claude Agent SDK places its **advancing** cache breakpoint precisely there (on the newest `tool_result`/assistant block each turn). So on the Anthropic ingress, agent traffic only ever caches the static system head; the growing conversation is re-billed every turn. (Confirmed on a customer's live GLM traffic: `cache_read` pinned at the ~12k head while `prompt_tokens` climbed to 33k, `cache_creation ≈ 0`.)

### What: preserve the marker as a single text-part array

Only messages that **carry a marker** change shape — everything else is byte-identical to today.

**Assistant (marked)** — incoming `/v1/messages`:

```jsonc
{ "role": "assistant", "content": [
  { "type": "text", "text": "Let me check that.", "cache_control": { "type": "ephemeral" } }
]}
```

```diff
- // before: flattened, marker lost
- { "role": "assistant", "content": "Let me check that." }
+ // after: preserved
+ { "role": "assistant", "content": [
+   { "type": "text", "text": "Let me check that.", "cache_control": { "type": "ephemeral" } }
+ ]}
```

**tool_result (marked)** — the common agent case:

```jsonc
{ "role": "user", "content": [
  { "type": "tool_result", "tool_use_id": "toolu_01", "content": "42 rows",
    "cache_control": { "type": "ephemeral" } }
]}
```

```diff
- // before: tool message, string content → marker lost
- { "role": "tool", "tool_call_id": "toolu_01", "content": "42 rows" }
+ // after
+ { "role": "tool", "tool_call_id": "toolu_01", "content": [
+   { "type": "text", "text": "42 rows", "cache_control": { "type": "ephemeral" } }
+ ]}
```

**Unmarked → unchanged** (still a plain string, both roles):

```jsonc
{ "role": "assistant", "content": "Let me check that." }
{ "role": "tool", "tool_call_id": "toolu_01", "content": "42 rows" }
```

**What the model receives:** the cache layer strips `cache_control` before the upstream call (as it already does for user/system blocks), so the upstream sees a single text part with no marker:

```jsonc
{ "role": "tool", "tool_call_id": "toolu_01", "content": [ { "type": "text", "text": "42 rows" } ] }
```

### Correctness invariant (unit-tested)

A single-text-part array with the marker stripped hashes **identically** to the plain-string form. That keeps the cache prefix stable as the marker advances off a block on the next turn (when it reverts to a string) — the prior write still matches, so the read lands. See `single_text_array_hashes_like_string_content` in `parse.rs`.

### Questions for reviewers

1. Was the flatten-to-string deliberate template-safety for a specific backend (esp. **sglang** chat templates on GLM/Qwen), or convenience? Array-form text-part content is OpenAI-spec-legal and the same shape our native `/v1/chat/completions` ingress already accepts — but I want to confirm it's safe end-to-end on our on-prem models.
2. Any objection to the single-text-part array on marked messages only?

PR #1263, opened 2026-07-13
feat: serve batch analytics from batch_aggregates + drop http_analytics index (COR-524 contract)

sejori:
## COR-524 — contract (PR 2 of 2)

Stacked on #1262 (the expand PR). Repoints the batch-analytics reads onto `batch_aggregates` and drops the reclaimed index. **Merge/deploy only after #1262 is fully rolled out and its backfill has run.**

### What's in this PR
- **`get_batch_analytics` + `get_batches_analytics_bulk`** now read the denormalized token/latency/cost columns from `batch_aggregates` instead of scanning `http_analytics` by `fusillade_batch_id`. `avg_duration_ms`/`avg_ttfb_ms` are reconstructed from the stored `sum`/`count` (SQL `AVG`, which ignores NULLs, is reproduced by counting only requests that reported the metric). A missing row = the old empty result (zeroes / `None`).
- **migration 118** — `DROP INDEX CONCURRENTLY IF EXISTS idx_analytics_fusillade_batch_id` (~1.6 GB), `-- no-transaction`.
- Tests repointed to populate `batch_aggregates` (the fold helper writes the read model).

> **Migration renumber (2026-07-13):** the drop is now **migration 118** (was 117). #1257 introduced `116_credits_service_tier_check`, so the COR-524 stack shifted up: #1262 expand = `117_batch_aggregates_analytics_denorm`, this PR = `118_drop_...`. The stack was rebased onto the current #1257 tip to make the migration sequence contiguous.

### Rolling-deploy gate on the index drop — decision: ship in-PR, mitigate with a quiet-window release
Unlike COR-516's drop, **this index's only readers are removed in this same PR**. If migration 118 runs on the first new pod's boot, old pods still running the `http_analytics` query would seq-scan ~130M rows on `/batches/{id}/analytics` until the rollout completes.

**Decision:** we keep the drop in this PR (migration 118) rather than splitting it to a follow-up, and **mitigate by releasing during a quiet window when load on `/batches/{id}/analytics` is effectively zero.** The hazard is a transient seq-scan only for the seconds/minutes of the rolling deploy, and only for requests actually hitting that endpoint — during the chosen window there are essentially none, so the pageserver-I/O risk is negligible.

Pre-flight before merging/deploying:
- Confirm via prod `pg_stat_user_indexes` that `idx_analytics_fusillade_batch_id` scans have gone flat after #1262's repoint is live (i.e. the read model is serving), and
- Deploy in the agreed low-traffic window; keep the rollout short so the old/new pod overlap is minimal.

(The stricter option — hold migration 118 to a separate release gated on full rollout — remains available if the window slips, but is not needed given the quiet-period plan.)

### Semantics note (carried from #1262)
`batch_aggregates` reflects **billed** successful requests (priced, `total_cost > 0`) rather than all `status 2xx` rows. ⚠️ This under-counts **free / zero-priced** models (they get no credit row, so they were never folded). Fix tracked on #1262 (fold the analytics aggregates off the newly-inserted `http_analytics` rows, not the billed set) so `total_requests`/tokens cover all 2xx again. For priced batch traffic the numbers are unchanged; `total_cost` stays the list price (`total_list_cost`).

### Retention: 404 when a batch's analytics have aged out
Because `batch_aggregates` is the durable read model (not subject to `http_analytics` retention), a **missing row** no longer means "in-flight" — it means the batch predates the read model and its raw rows have aged out. `get_batch_analytics` now returns `Option`, and `GET /batches/{id}/analytics` returns **404 ("analytics no longer available")** rather than a misleading all-zero payload. The dashboard only requests analytics for *completed* batches (`hooks.ts`), so this never fires on a genuinely-pending batch. (Edge case: an all-non-2xx batch also has no row and thus 404s — acceptable; "no analytics available" is truthful there too.)

This is the agreed handling for the persisted-data-vs-retention gap: we do **not** build a per-request durable store; the UI shows "analytics for this batch are no longer available" on the 404. Note the sibling per-request enrichment in `batch_requests.rs` (`GET /batches/requests[/{id}]`) still joins `http_analytics` by `fusillade_request_id` and would need the same 404 treatment in a follow-up if that drill-down must survive retention.

### Tests
The 8 `get_batch_analytics` / bulk tests pass against the read model; clippy clean; `cargo sqlx prepare --check` clean; offline build green. **Note:** `.sqlx` offline cache must be regenerated (`cargo sqlx prepare --workspace`) after the rebase before this is CI-green.


sejori:
## Ready to deploy — scheduled for **4:30 (quiet window)** with a final backfill pass

### Backfill prerequisite: done + validated (set-based)
The original per-batch backfill was infeasible at prod scale — ~500k random `http_analytics` index probes ≈ **30 batches/min (~11 days)** and it saturated Neon `PS_ReadIO` against live traffic. Rewritten in this PR (`perf(backfill): rewrite batch backfills as set-based single-pass`) to **one parallel sequential scan + hash-aggregate** over `http_analytics` grouped by `fusillade_batch_id`, then two bulk `UPDATE`s.

| env | batches backfilled | time | correctness (sample vs per-batch recompute) |
|---|---|---|---|
| **prod** | 490,171 | ~39s (13s scan + 22s update, bumped compute) | 0 mismatches / 300 |
| **staging** | 490,196 | ~4.5min (unbumped compute) | 0 mismatches / 350 |

`batch_aggregates.service_tier` is fully populated on both; temporary helper indexes cleaned up.

### Deploy plan (at 4:30, low-traffic)
1. **mig 118 rolling-deploy gate** ⚠️ — mig 118 drops `idx_analytics_fusillade_batch_id`. During this PR's rollout, old pods still read `http_analytics` by `fusillade_batch_id` (pre-repoint `get_batch_analytics`), so deploy in the quiet window and confirm the index is idle (`pg_stat_user_indexes.idx_scan` flat) before/at the drop.
2. **Final backfill pass alongside the deploy** — sweeps the *quiescence tail*: historical batches that were `<25h` idle at the initial run and have since crossed the 25h guard. Post-deploy batches are covered by the go-forward fold; only recently-touched historical batches need this pass. Now ~seconds.

### Side benefit
The set-based backfill uses a **sequential scan (not the `fusillade_batch_id` index)**, and the service_tier backfill reads `fusillade.batches` — so **neither backfill depends on the index mig 118 drops**. The COR-524 "hard split" ordering constraint (backfill must finish before the index drop) is effectively gone.

sejori:
### Migration 119 (index drop) rolling-deploy gate — deploy plan

Ran the gate against production. **Refining the risk note in the PR body:** the `idx_analytics_fusillade_batch_id` traffic is *not* occasional `GET /batches/{id}/analytics` views — it's the batches **list** via `?include=analytics`, which both the admin dashboard and the customer-facing frontend request on every list fetch and **auto-poll** (~10s interval). The pre-repoint `get_batches_analytics_bulk` runs `FROM http_analytics WHERE fusillade_batch_id = ANY($1)` — i.e. one index scan per batch, per poll.

Measured live on the primary: **continuous ~290 index scans/min**, `last_idx_scan` always current. This PR repoints both `get_batch_analytics` and `get_batches_analytics_bulk` onto `batch_aggregates`, so scans fall to ~0 once the repoint is fully rolled out — but during the rolling deploy, old pods keep polling `http_analytics`, and after the drop each poll seq-scans the full table (cold-read I/O saturation risk on the primary).

**Plan** (chosen over splitting the drop into a follow-up release):

1. **Deploy in tomorrow's low-traffic window (early UK morning)** — fewest open Batches sessions polling.
2. **Actively cancel the pre-repoint analytics seq-scans** on the primary during the old→new pod overlap, so straggler polls landing on old pods can't sit and saturate I/O. Target **only** those queries — *not* all activity, which would also kill the batcher/fold/other live traffic:
   ```sql
   -- inspect the matches first, then cancel; run in a short loop until old pods drain
   SELECT pg_cancel_backend(pid), now() - query_start AS runtime, left(query, 120)
   FROM pg_stat_activity
   WHERE state = 'active'
     AND datname = current_database()
     AND query ILIKE '%from http_analytics%'
     AND query ILIKE '%fusillade_batch_id = any%'
     AND pid <> pg_backend_pid();
   ```
3. **Keep the rollout short** (high surge / fast old-pod termination) to minimise the overlap window.

**Residual risk:** cancelling is reactive — a scan does some I/O before it's cancelled, and each cancel surfaces a transient error on that list poll (clients retry). We're accepting this small, actively-managed window instead of the zero-risk alternative (ship the repoint here, confirm scans are flat in prod, then drop the index in a follow-up).

**Pre-/at-drop confirmation:** `idx_scan` on `idx_analytics_fusillade_batch_id` flat + `last_idx_scan` going stale ⇒ every pod is repointed and nothing reads the index → safe to drop.

_(Note: this migration was renumbered 118 → 119 after `118_add_reasoning_translation` landed on `main`.)_

PR #1273, opened 2026-07-14
perf: add covering indexes for TTFT analytics

pjb157:
## Summary

- add concurrent partial covering indexes for successful streaming TTFT window queries
- order each index by the computed first-byte event time
- cover queued-request correlation fields without widening the indexed population

## Verification

- applied migrations 1 through 119 to a clean PostgreSQL database
- confirmed both indexes are ready and valid
- confirmed realtime and queued-request 30-day scans plan as index-only scans


pjb157:
Closing because TTFT reporting will be computed outside the application database schema.

Issue #139, opened 2025-11-19
Add toast notifications for copy actions


fergusfinn:
## Description
Add toast notifications when users copy API keys or other values to improve feedback.

## Problem

**In `ApiKeys.tsx` (lines 141-153):**
- Copy button shows checkmark for 2 seconds
- No persistent indication after timeout
- User might click multiple times if they don't notice icon change
- No confirmation if copy failed

## Solution

Already using `sonner` toast library! Just need to add calls:

```typescript
import { toast } from "sonner";

const copyToClipboard = async (text: string) => {
  try {
    await navigator.clipboard.writeText(text);
    setCopiedKey(text);
    toast.success("API key copied to clipboard"); // ✅ Add this
    
    setTimeout(() => {
      setCopiedKey(null);
    }, 2000);
  } catch (err) {
    console.error("Failed to copy to clipboard:", err);
    toast.error("Failed to copy API key"); // ✅ Add this
  }
};
```

## Files to Change
- `dashboard/src/components/features/api-keys/ApiKeys/ApiKeys.tsx`

## Other Copy Actions
Search for other copy actions that could benefit:
- Copying request IDs
- Copying endpoint URLs
- Any other clipboard operations

## Design
- Success: Green toast with checkmark
- Error: Red toast with X icon
- Duration: 2-3 seconds (default)
- Position: Bottom-right (current toast config)

## Status
⏳ Pending

PR #1275, opened 2026-07-14
feat(usage-e): denormalize fusillade_request_id onto credits + backfill (expand)

sejori:
## Usage E — expand half (COR-524 follow-up)

Makes per-request **cost** in the responses view durable by keying it to the request on the credits ledger, so it survives http_analytics retention (COR-509). This PR is the **expand** half; the **read repoint** ships in the stacked contract PR **#1278** (after the backfill).

### What's here
- **migration 119** — `credits_transactions.fusillade_request_id UUID` (nullable → metadata-only).
- **migration 120** — partial `CREATE INDEX CONCURRENTLY (fusillade_request_id, seq DESC) WHERE fusillade_request_id IS NOT NULL` (`-- no-transaction`), keyed for the contract's `DISTINCT ON … ORDER BY seq DESC` reader.
- **batcher** — sets `fusillade_request_id` on the credits insert.
- **`scripts/backfill_credits_fusillade_request_id.sh`** — keyset-by-PK backfill from http_analytics (`ha.id = source_id::bigint`, guarded by CASE), Neon-tested at prod scale.

### No read change here
The responses view still reads cost from http_analytics in this PR — so merging it does **not** blank any historical cost. See the coordination comment below for the full merge/backfill/deploy sequence with #1263.


sejori:
## Merge / backfill / deploy coordination (with #1263)

Tested the migration + backfill on the prod-scale DB branch (**28.5M** credits, **131M** http_analytics). Answers to the open questions:

### 1. Does it rely on any index #1263 drops? — No.
#1263 drops **`idx_analytics_fusillade_batch_id`** (http_analytics · `fusillade_batch_id`). Nothing in Usage E touches that index:
- **Backfill** joins http_analytics by **`ha.id = source_id::bigint` → `http_analytics_pkey`** (the PK, always present). The `id::text` form the COR-507 notes flagged as heavy is deliberately avoided, so the already-dropped `idx_http_analytics_id_text` doesn't matter either.
- **Contract read** (#1278): cost from **`credits_transactions` via the new index 120**; tokens still from http_analytics by **`fusillade_request_id` → `idx_analytics_fusillade_request_id`** (a *different* index, kept).

So Usage E and #1263 are independent. **One ordering constraint:** migrations 119/120 must apply *after* #1263's 118, so **merge #1263 first** (else sqlx sees 118 as an out-of-order migration below an applied 119).

### 2. Does the backfill need helper indexes to run effectively? — No.
It keyset-paginates by the **credits PK** (`WHERE id > $last ORDER BY id LIMIT n`) and probes http_analytics by **its PK**. Measured plan = three nested-loop PK index scans, **zero seq-scans**. (The original seq/`created_at`-range sweep was the trap: `seq` has no standalone index and the planner won't range-scan `created_at` with the extra filters, so each batch seq-scanned all 28.5M rows ~90s → hours. Fixed in the latest commit.) No temporary index to build or drop.

### 3. Will merging leave the app broken until backfill runs? — Not this PR; that's why it's split.
- **#1275 (expand) — safe to merge/deploy on its own.** It only *adds* the column, sets it on new credits, and ships the backfill script. The responses view still reads cost from http_analytics here, so no historical cost is blanked.
- **#1278 (contract) — must NOT deploy before the backfill.** It moves the cost read to `credits.fusillade_request_id`, which is NULL for pre-deploy rows until backfilled → old responses would show `—` for cost. Hence the expand→contract split.

### 4. Is a follow-up PR needed to migrate the API after backfill? — Yes: **#1278** (stacked on this).

### Recommended sequence
1. Merge **#1263** → deploy → (its own backfill/index-drop per that PR).
2. Merge **#1275 (this)** → deploy fully so the batcher populates `fusillade_request_id` on new credits.
3. Run **`backfill_credits_fusillade_request_id.sh`** (keyset, ~20k/batch, resumable, off-peak) — **before COR-509 prunes http_analytics**; rows already aged out can't be linked. If full history isn't needed, start from a recent id to scope it to the retention window.
4. Confirm the link is populated, then merge **#1278** + **app-doubleword-private #139** → deploy. Cost is now durable end-to-end.

### Performance note
Full-scale backfill is inherently a multi-hour, batched, resumable job (one row per usage credit: a PK probe into http_analytics + a write) — same class as the COR-507 credits backfill. The keyset algorithm removes the pathological per-batch seq-scan; the residual cost is Neon pageserver I/O for ~28.5M cold reads/writes, best run off-peak on the primary.

PR #1332, opened 2026-07-22
feat: add service_tier label to gen_ai metrics and extend TTFT/TPOT buckets

fergusfinn:
## What

- Adds a `service_tier` label (`realtime` / `flex` / `async` / `batch`) to all four `gen_ai_*` metric families, reusing the same `compute_service_tier` derivation the credits ledger already uses (now `pub(crate)`), so metrics and ledger cannot disagree.
- Extends histogram buckets: TTFT gains `15, 20, 30, 45, 60, 90, 120`s; TPOT gains `5, 10`s.

## Why

Per-tier latency measurement (SLO adherence dashboards) currently has to reconstruct the tier from `request_origin` × `batch_sla`, which cannot distinguish flex from 1h batch traffic. The bucket tails let compliance and p99 be read against per-tier bounds (TTFT 10s/60s, inter-token 50ms/1s) — those bounds are deliberately exact bucket edges, since a compliance ratio `bucket{le="X"}/count` is only exact at an edge.

## Notes

- Label cardinality: ×4 worst case, sparse in practice (tier is a function of existing label values, so no new label-combination blowup beyond the split it introduces).
- Additive: existing series names, labels, and dashboards are unaffected; bucket extension appears as new `le` series after deploy.
- No behavior change; metrics-only.

## Verification

- `cargo test -p dwctl --lib -- request_logging:: metrics::` — 138 passed, including a new test pinning all four tier mappings.
- Post-deploy check: `sum by (service_tier, request_origin, batch_sla) (increase(gen_ai_server_request_duration_seconds_count[1h]))` should show only legal combinations (`realtime` ↔ empty `batch_sla`; `flex`/`async`/`batch` ↔ `fusillade`).


dwctl/src/request_logging/batcher.rs
pjb157, line 1681: I guess this was already here but I don't like the mix of service tier which is the openai compatible parameter with our own product logic

pjb157: Can we change the name to something doubleword specific to save my poor brain

pjb157 (approved):

PR #1423, opened 2026-08-03
feat(users): expose the verified flag on the user API

sejori:
## Why

dwctl has tracked `users.verified` for a while — the payment-success paths set it, the new setup-mode card verification (#1421) sets it, and the onwards sync reads it to choose between the verified and unverified rate-limit tiers.

It has never been readable over the API.

The onboarding ZDR step needs it. Zero data retention is gated on the account having proven it can pay, and today the SPA has to infer that from `has_payment_provider_id`. **That's the wrong signal** — it only says a customer record exists at the provider, and `enable_auto_topup` creates one purely to check for a billing address, with no card attached. Gating ZDR on it would let an unverified account switch on zero data retention.

## What

Adds `verified` to `UserResponse`, plumbed from the `users` row that already carries it.

Organizations are users too and can be verified in their own right — setup checkout verifies the *resolved billing target*, which is the org when you're acting as an org admin — so the three `Organizations` queries now select it as well.

Purely additive: one new field on a response body, no behaviour change.

## Test plan

- [x] `cargo test -p dwctl` — **1936 passed, 0 failed**
- [x] `just lint rust` — clean (fmt, clippy, ZDR payload guard, `sqlx prepare --check`)
- [x] New test asserts the flag reaches `/users/current` **and flips there** after `set_verified` — a value that never leaves the database is precisely the bug being fixed here, so the test walks the transition rather than just reading the end state
- [x] Existing `test_get_current_user_info` now asserts a fresh account reports `verified: false`

## Follow-up

Once this is out, the SPA's onboarding billing step should move off `has_payment_provider_id` and onto `verified` for anything that means "this account can pay". Tracked alongside app-doubleword-private#167.


sejori:
## Addressed Copilot's replication-lag point — it was a real hole

Copilot flagged that `GET /users/{id}` (including `current`) reads from the replica, so `verified` can come back stale right after a verification write. That's not theoretical here, it's the whole point of the flag:

1. `PATCH /payments/{id}` sets `verified` on the **primary**
2. the SPA immediately refetches this endpoint to decide whether the account earned zero data retention
3. a lagging replica answers `verified: false`
4. **ZDR is silently not applied** — the account keeps standard retention while the UI has already moved on to the handoff

CLAUDE.md calls this out directly as the inter-endpoint consistency trap, and notes Neon shows 100ms–1000ms of lag under load. Exactly the window between those two calls.

`get_user` now uses the primary pool. Costs it a single-row primary-key lookup. `list_users` stays on the replica — nothing chains a write into a list view.

The new test deliberately **does not poll or retry**: the point is that the very next read is already correct, and a polling test would pass just as happily against the broken version.

## Also rebased onto latest main

Picks up #1424. Note this replaced the `Merge branch 'main'` commit that was on the branch — I checked the diff against main first and it was byte-identical to mine, so nothing was lost.

## Test status

- `just lint rust` — clean
- `cargo test -p dwctl` — **1930 pass**

10 failures remain locally, all of which **also fail on `main` unchanged** and are network-bound in this sandbox: the 9 new organization email-change tests (they hit `mx_lookup` at `organizations.rs:120`) plus `anthropic_messages_blocking_end_to_end`. CI has network, so it's the arbiter — flagging so the local numbers aren't mistaken for a regression.

PR #1429, opened 2026-08-04
feat(payments): scaffold invoice billing for enterprise accounts

sejori:
**Draft** — groundwork for billing enterprise accounts by monthly invoice instead of charging a card. The branching and the provider call are here; the Stripe-side billing cycle that actually raises the invoice is not.

## Why accrual, not "create and send an invoice"

My first pass raised one invoice per top-up and emailed it immediately. That's wrong for this feature: auto top-up can fire many times a month, so it would put **an invoice per firing** through the customer's AP department — the exact thing "monthly invoicing for PO processing" exists to avoid.

So `accrue_invoice_item` creates a **pending** Stripe invoice item — no `invoice` field — which Stripe parks on the customer and sweeps onto their next invoice. One invoice per period, covering everything accrued in it, raised and emailed by Stripe.

## This is post-pay, and that matters

Credits are granted **at accrual, before any money arrives**.

The tempting alternative — grant on `invoice.paid` — is self-defeating: an invoice-billed account would sit at a zero balance all month with auto top-up unable to do anything about it, because the top-up it just performed hasn't been paid for yet.

So the account is genuinely being **extended credit on terms**. That's why `invoicing_enabled` is set by us after approval rather than self-served.

## What's here

| | |
|---|---|
| **Migration 129** | `users.invoicing_enabled`, default `false`. Orgs are rows in `users`, so billing an org works the same way. |
| **`accrue_invoice_item`** | on `PaymentProvider`, implemented for Stripe and the dummy provider |
| **Auto top-up branch** | accrues + credits, then returns |
| **Manual top-up branch** | refuses with `409` |

**The auto top-up branch sits before the payment-method lookup deliberately.** These accounts may have no card at all, and everything below it — soft/hard decline backoff, "auto top-up disabled" emails — describes a failure mode that cannot happen here. Non-payment of an invoice is collections on Stripe's dunning schedule, not a card decline.

**Manual top-up refuses rather than accruing.** Sending an invoice-billed account to Checkout would either fail for want of a card, or take a payment for credits their monthly invoice is also going to cover — billing them twice for one top-up.

## Not done — needed before this is a feature

- [ ] The Stripe **subscription** (or equivalent) whose billing cycle raises, finalises and emails the monthly invoice. Stripe won't do this for a bare customer: pending items sit indefinitely until *something* creates an invoice. With `auto_advance: true` on the generated invoice, Stripe then handles delivery, reminders and receipts with no scheduler on our side.
- [ ] Suspension / collections policy on non-payment. Right now an unpaid account keeps accruing.
- [ ] A way to set the flag (currently SQL only — deliberate, given it extends credit).
- [ ] Whether the ledger should record the invoice item ID for reconciliation.

## Test plan

- [x] `cargo test -p dwctl` — **1942 passed, 0 failed**
- [x] `just lint rust` — clean
- [x] New: an invoice-billed account is credited **without touching a card** (the test uses the dummy provider's always-declines customer, so a leak into the card path would fail it); the flag defaults false so existing accounts keep the card path unchanged; manual checkout is refused *with a reason*

Inert until the column is set, so it's safe to land ahead of the remaining work.


sejori:
Marking ready so it picks up a review — this is **still not for merging**. It's the invoice-billing scaffold: the branching and provider call are here, the Stripe billing cycle that actually raises the monthly invoice is not (see the unchecked list in the description).

PR #1430, opened 2026-08-04
feat(organizations): replace silent domain auto-join with join requests

sejori:
Phase 1 of making domain-based org membership deliberate. Unblocks the **Requests** and **Invitations** tabs of doublewordai/app-doubleword-private#134.

## The behaviour this replaces

Until now, `auth::current_user` did this on first login with a company email domain:

```rust
Ok(Some(org)) => {
    // Org exists — add user as member
    org_repo.add_member(org.id, user.id, "member")
}
```

**No approval, no notice to either party.** Anyone who could receive mail at a domain landed inside the organization that owns its billing account, and the org found out by noticing a new name in the members list.

Sharing an email domain is evidence that someone *probably* belongs — not proof that they do. They now land in a queue and an owner or admin decides.

## Why no new table

Join requests are the mirror image of invites, so they reuse the same membership lifecycle:

| status | meaning |
|---|---|
| `active` | current member |
| `pending` | **org invited someone**, not yet accepted |
| `requested` | **someone asked**, org hasn't decided |

Both non-active states become `active` on acceptance. Both are already covered by `UNIQUE (user_id, organization_id)`, the membership-type trigger and the existing indexes — so repeat logins can't pile up duplicates, and that's tested.

The migration is one CHECK constraint plus a partial index. No new table, no new trigger, no second lifecycle to keep in sync.

## Decisions worth reviewing

**Gated on *managing* the org, not belonging to it.** The queue names people who share your email domain; ordinary members have no business seeing it. There's a test asserting a plain member gets `403` where the owner gets the list.

**Declining deletes the row rather than tombstoning it.** A permanent `declined` record would silently block that user forever via the unique constraint. Tested by declining and then asking again.

**Resending mints a new token.** The original is only stored as a hash, so it *can't* be recovered to send again — a resend necessarily rotates it. That also kills the old link, which is what you want if the first went astray. Tested by asserting the stored hash changes.

**`list_members` now excludes `requested`** so the new status can't leak into the members or invitations views.

## Deliberately not in scope

Auto-**creation** of an org for an unclaimed domain is unchanged — first signup from a new domain still creates and owns it. Gating that needs domain ownership proof (DNS TXT), which is phase 2.

Worth flagging the residual risk that leaves: someone who claims `acme.com` first could approve real Acme staff into *their* org. That risk exists today and is strictly worse today, since it's currently auto-join with no prompt at all — but it's the reason phase 2 matters.

## Test plan

- [x] `cargo test -p dwctl` — **1947 passed, 0 failed**
- [x] `just lint rust` — clean
- [x] The old `test_proxy_header_signup_joins_existing_org` is rewritten to assert the *new* contract: `get_user_org_role` returns `None` (the assertion that would have failed before), the request is queued, and the members list is unaffected
- [x] New: repeat logins don't duplicate; approve makes an active member and is a no-op the second time; decline lets the user re-request; decisions are scoped to their own org (owner of A gets `403` on B's queue and B's request stays outstanding); resend rotates the token

### DIFF
--- a/.sqlx/query-16161654b04e7faa5eacd3f40e6a4fad922fc28f8d87cbdb8b124414446ed6d5.json
+++ b/.sqlx/query-16161654b04e7faa5eacd3f40e6a4fad922fc28f8d87cbdb8b124414446ed6d5.json
@@ -1,0 +1,71 @@
+{
+  "db_name": "PostgreSQL",
+  "query": "\n            INSERT INTO user_organizations (user_id, organization_id, role, status)\n            VALUES ($1, $2, 'member', 'requested')\n            RETURNING id, user_id, organization_id, role, status, created_at,\n                      invite_email, invited_by, expires_at\n            ",
+  "describe": {
+    "columns": [
+      {
+        "ordinal": 0,
+        "name": "id",
+        "type_info": "Uuid"
+      },
+      {
+        "ordinal": 1,
+        "name": "user_id",
+        "type_info": "Uuid"
+      },
+      {
+        "ordinal": 2,
+        "name": "organization_id",
+        "type_info": "Uuid"
+      },
+      {
+        "ordinal": 3,
+        "name": "role",
+        "type_info": "Varchar"
+      },
+      {
+        "ordinal": 4,
+        "name": "status",
+        "type_info": "Varchar"
+      },
+      {
+        "ordinal": 5,
+        "name": "created_at",
+        "type_info": "Timestamptz"
+      },
+      {
+        "ordinal": 6,
+        "name": "invite_email",
+        "type_info": "Varchar"
+      },
+      {
+        "ordinal": 7,
+        "name": "invited_by",
+        "type_info": "Uuid"
+      },
+      {
+        "ordinal": 8,
+        "name": "expires_at",
+        "type_info": "Timestamptz"
+      }
+    ],
+    "parameters": {
+      "Left": [
+        "Uuid",
+        "Uuid"
+      ]
+    },
+    "nullable": [
+      false,
+      true,
+      false,
+      false,
+      false,
+      false,
+      true,
+      true,
+      true
+    ]
+  },
+  "hash": "16161654b04e7faa5eacd3f40e6a4fad922fc28f8d87cbdb8b124414446ed6d5"
+}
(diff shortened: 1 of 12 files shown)


sejori:
Thanks — four of the five are real and are fixed in `8eb69c5`. One doesn't hold; detail below.

## Fixed

**The displaced doc comments (`organizations.rs:745` + the suppressed `:771`) — my bug, and the worst of the batch.** I inserted both new items immediately *above* an existing declaration instead of below the preceding one, so each stole the doc comment belonging to the item underneath.

`refresh_invite_token` ended up wearing `cancel_invite`'s doc **and** its `#[instrument]`, on top of its own. Two instrument attributes compose, and the stolen one only skips `self` — so `token_hash` would have been recorded as a span field. It's a hash rather than a usable token (you'd still need the preimage to redeem an invite), so this isn't exploitable, but it was never meant to be logged. Meanwhile `cancel_invite` lost its documentation and its tracing span altogether, which is the part I'd actually have missed in production.

Same mistake in `models/organizations.rs`: `ApproveJoinRequestRequest` took "Organization member details" from `OrganizationMemberResponse`, which reaches the generated OpenAPI schema.

**`approve_join_request` advertised 200, returns 204** — corrected. `decline` and `resend` were already right, so this was just the one.

**`OrganizationMemberResponse.status` docs** — now cover `requested` alongside `active` and `pending`.

## Not valid: the warn-on-repeat-login concern

> Logging it at `warn` on every subsequent login is likely to create noisy/false-positive warnings in production.

The duplicate insert can't happen on repeat logins, because the whole auto-org block sits inside `if was_created {` (`current_user.rs:173`) — it only runs when the user row is first created. Logins two onward never reach it.

`test_proxy_header_repeat_login_does_not_duplicate_join_request` passes for exactly that reason: three logins produce one request, and the second and third don't touch this code at all rather than being rejected by the constraint.

The only way to reach the unique violation is a genuine race between two concurrent first-logins for the same user, which is both vanishingly rare and worth a warning — if the insert fails there, the user silently isn't queued and nobody finds out. So I've left the level as-is.

## Verification

`cargo test -p dwctl` — **1947 passed, 0 failed**. `just lint rust` clean.

PR #1479, opened 2026-08-13
feat: add privacy-aware request logging

pjb157:
## Summary

- add independently configurable request and response header allowlists for persistent logging
- derive an opaque log subject from the active authenticated-key policy cache, reject unavailable attribution without database work, and strip its internal carrier before upstream dispatch
- keep HTTP headers out of serialized request content and preserve body scrubbing across single, abandoned, and batch callbacks
- coalesce authentication-policy cache notifications so closely spaced mutations receive a trailing refresh
- validate request-logging configuration on file and programmatic startup paths

## Dependency

- depends on doublewordai/outlet-postgres#84, pinned to its reviewed commit until released

## Verification

- `just lint rust`
- `just test rust`
- focused active/deleted-key, cache-coalescing, spoofing, carrier-stripping, serializer, configuration, and batch-callback tests
- independent review found no critical or important issues


pjb157:
Closing after review: these logging changes are not required for the retention capability and are intentionally excluded.

PR #1482, opened 2026-08-13
feat: add durable data erasure

pjb157:
## Summary

- make account erasure a durable, transactionally enqueued workflow with completion receipts and indefinite retry
- add permanent creator replay guards that serialize with in-flight writes, bounded content draining, absence verification, and residual parent scrubbing
- erase optional request-capture stores through generic subject indexes and fail closed when required indexes or stores are unavailable
- remove credentials, access grants, request-level identifiers, and other account-scoped artifacts while preserving non-content ledger integrity
- expose aggregate completion/retry metrics without placing raw subject identifiers in queue payloads or logs

## Dependencies

- stacked on #1481
- consumes the bounded subject-erasure primitives from doublewordai/outlet-postgres#85

## Verification

- `cargo test --workspace --profile ci -- --test-threads=4`
- `just lint rust -- -D warnings`
- `cargo fmt --all -- --check`
- `git diff --check`


pjb157:
Closing after an independent scope and production-readiness audit. The existing account-deletion flow plus bounded age retention and operational verification covers the retained requirement without this cross-store workflow. This implementation also conflicts with immutable ledger protections, changes queued-job payload compatibility, and introduces disproportionate startup and write-path risk. Narrow credential revocation and queue-reliability fixes should be handled separately.

PR #1492, opened 2026-08-14
feat(recompute): add the read-only dry-run endpoint

sejori:
`GET /admin/api/v1/usage-recompute`. Replays a corpus of already-billed requests through the live serializer, re-prices them, and returns stored vs recomputed usage per request.

**Read-only by construction** — the handler is given `state.db.read()` and holds no write handle, so it cannot mutate even by mistake. Applying a correction is a separate human-run sequence, documented in the internal `usage-recompute-repair` runbook.

Stacked on #1491 (reasoning/total), which the report shape depends on. **Merge together with #1494**: prod verification found that the multiplier inference this PR shipped with re-prices dwctl-cached rows wrongly, and #1494 carries the fix (resolving the `model_cache_tariffs` version valid at the request's pricing time, exactly as the live batcher does) plus the endpoint's integration tests. This PR is kept as-authored to preserve review history; the stack is the unit that ships.

## Shape

- **`source.rs`** — one bounded scan joining `http_analytics` to the fusillade bodies. Deliberately *not* per-row lookups: that pattern saturated the Neon pageserver during the July remediation and had to be killed after 14 minutes.
- **`report.rs`** — the JSON document: every usage field on both sides, plus `Evidence` (`BodyFrame` / `ColumnsOnly` / `NotReplayable`).
- **`recompute_corpus`** — source → engine → pricing → report.
- **handler** — validates the predicate (window ≤ 31 days, limit ≤ 25k) and rejects an unknown `flat_creation_tier` rather than silently defaulting.

## Three decisions worth your attention

**Re-pricing uses the per-token rates stored on the analytics row**, not a freshly-resolved tariff. Re-resolving would silently re-base a correction onto pricing that changed *after* the request was served.

**Cache multipliers were inferred from the row's own evidence in this PR — superseded in #1494.** Measured against healthy prod traffic, the inference invented multipliers (config defaults: read ×0.1, write ×1.25) where the live tariff was read ×0.8, write ×1.0, reporting a negative "correction" on healthy rows. The tariff ledger is temporal, so #1494 resolves the version valid at the request's pricing time instead — the exact multipliers the live path billed with.

**Un-replayable rows contribute their *stored* figures to the recomputed totals**, so `net_correction` is the real delta rather than one inflated by rows nobody could verify. They're reported as `ColumnsOnly`/`NotReplayable` rather than dropped — a corpus with silent omissions is worse than one that says what it couldn't do.

## Tests

Three at corpus level against a real database:

- **healthy traffic → `rows_changed = 0`, `net_correction = 0`** — the property the whole feature rests on, now exercised through the real query rather than a synthetic fixture
- the August incident end to end: 565 → 18,734, creation recovered from the **flat** field, `cache_tier_inferred` set
- a body-less row reported as `ColumnsOnly`, not dropped, and not counted as a change

Handler-level integration tests (permission gate, predicate validation, both incident shapes over HTTP) are on #1494 with the fixes they exercise.

## Housekeeping

Drops the module-wide `allow(dead_code)` now the engine has a real consumer. Only `fold_render` keeps a scoped one — the tokenizer path still has no caller, and its tests encode the precedence rules that are the hard part.


sejori:
Local verification now complete, updating the note in the description:

- `cargo test -p dwctl --lib` — **2019 passed, 0 failed**
- `just lint rust` — clean, including `sqlx prepare --check` (the new corpus query needed a `.sqlx` regeneration, which is in the commit)

sejori:
## ⚠ BLOCKED — do not merge

Verification against prod found a defect in `recompute_from_stored_response` that affects this PR.

**For dwctl-cached traffic the raw fusillade body does not carry dwctl's injected cache fields.** The recompute carries the split through from that body, so it reads `cache_creation = 0` where the row correctly holds 26,862, and re-prices on a split that is not there. On healthy recent traffic it produced a **negative net correction** — it claimed we overcharged, which should be impossible given the zero-delta property this feature rests on. One row recomputed to `prompt_tokens = 0` outright.

This is the same class as the incident that started this work: reading a body that does not carry what you assume it carries.

Fix and context: `recompute-split-source-handover.md` in the workspace root. Neither this PR nor #1494 should merge until it is corrected and the verification shows a non-negative net correction on healthy traffic.

PR #1494, opened 2026-08-14
feat(recompute): reconstruct the cache split as of a past instant

sejori:
Stacked on #1491 → #1492; this PR completes the stack and is where prod verification of the whole feature happened. **The three PRs merge together, in order.**

## What it does

**Historical cache-split reconstruction.** Adds `HistoricalIndex`, a `CacheIndex` that answers *"was this prefix live at `at`"* instead of *"is it live now"*, plus a `principal` override on `ClassifyRequest`. The **serving classifier runs against historical data unchanged** — same parse, same breakpoints, same hashes, same render — so a reproduced row is evidence the serving pipeline was right, not merely self-consistent. Wired into the report as `reconstructed_cache` per row (split, render total, fidelity, agreement with the stored row).

Writes are refused, not ignored: a reconstruction that wrote to the index would corrupt the evidence for every later run. A historical row's principal is supplied directly because its api key may since have been rotated.

## Fixes from verifying against prod (the reason the stack was held in draft)

Verification on healthy prod traffic initially reported every row changed and a negative net correction — the outcome the design forbids. Root causes, all fixed here with regression tests, all re-verified against prod afterwards:

1. **Invented cache multipliers** (the money bug). Re-pricing used config-default multipliers wherever a row recorded cache tokens. `model_cache_tariffs` is temporal, so the version valid at the request's pricing time (batch-created-at for batch traffic) is exactly resolvable — now done through `pricing::lookup_cache_tariffs`, shared with the live batcher so the two cannot drift. A row whose tariff history no longer resolves re-prices at list rate and the report warns those deltas must not be applied.

2. **Raw-usage fallback, on the live path too.** Some upstreams answer 200 with a full usage object in a shape the typed parse cannot represent — OpenRouter emits `finish_reason: "error"`, Dynamo omits the message `role` — and the untagged parse falls to `Other`, reading zero. On the live path this billed real usage at nothing and wrote the impossible `prompt < read + creation` analytics signature (measured: 149 rows / 3.75M prompt tokens / ~$7.09 for one user in 10 days, still being written this week). The fallback lives in the serializer, shared verbatim by live billing and the replay; it fires only when the typed parse produced no counts at all.

3. **A usage-less body now refuses to recompute** (`NoUsage` → `rows_not_replayable`) instead of reporting a healthy zero. Absence of evidence is not a measurement: against a stored zero it certified the July incident as healthy, and against a stored non-zero it would have proposed zeroing out real usage.

4. **Degraded classify outcomes are flagged and refused as evidence.** The serving classifier deliberately degrades to a zero split when tokenizer-svc is unreachable — correct for serving, poison for verification: a dead port-forward turned a healthy 25-row corpus into 25 confident false disagreements. `ClassifyOutcome.degraded` marks infrastructure-failure zeros; reconstruction refuses them, so the row carries no `reconstructed_cache` rather than a fake one.

5. **Cost compares at the stored column's scale.** `http_analytics.total_cost` is numeric(12,8); the stored side was rounded by the column on insert. Comparing unrounded manufactured 194 phantom sub-cent "changes" on a healthy 400-row corpus.

Also here: integration tests for the `/usage-recompute` endpoint (permission gate, predicate validation, both incident shapes end to end over HTTP), and `RECOMPUTE_END` on the verify harness so a corpus can target an incident cluster.

## Verified against prod (read-only), 2026-08-14

- 25 healthy dwctl-cached rows (glm-5.2, heavy cache traffic): **0 changed, net correction exactly 0**, 24/25 reconstructions agreed with the stored split with render totals matching stored prompts exactly; the one disagreement is the documented pre-cutover over-approximation, flagged approximate.
- 400 healthy rows on a second model with a sub-8dp tariff: **0 changed, net 0** (194 phantom changes before fix 5).
- A corpus containing the `finish_reason: "error"` rows: the broken rows recompute to the body's real counts at tariff prices (stored 0 tokens / $0 → 29,366 prompt / $0.00589) while 998 healthy neighbours propose zero change.

## Bounds, carried in the output

- **Retention** — entries all still exist today; the sweeper's grace will eventually put older corpora out of reach (the report warns).
- **Fidelity** — pre-cutover entries over-approximate liveness (a past creation can read back as a live read); marked `approximate`. Resolved by the episode-per-row work.


sejori:
## ⚠ BLOCKED — do not merge

Verification against prod found a defect in `recompute_from_stored_response` that affects this PR.

**For dwctl-cached traffic the raw fusillade body does not carry dwctl's injected cache fields.** The recompute carries the split through from that body, so it reads `cache_creation = 0` where the row correctly holds 26,862, and re-prices on a split that is not there. On healthy recent traffic it produced a **negative net correction** — it claimed we overcharged, which should be impossible given the zero-delta property this feature rests on. One row recomputed to `prompt_tokens = 0` outright.

This is the same class as the incident that started this work: reading a body that does not carry what you assume it carries.

Fix and context: `recompute-split-source-handover.md` in the workspace root. Neither this PR nor #1494 should merge until it is corrected and the verification shows a non-negative net correction on healthy traffic.

PR #1514, opened 2026-08-25
feat(dashboard): explain what the region indicator means

sejori:
The dashboard header has always shown `Region: <name>` when `metadata.region` is configured — but nothing told the user what it implied. This adds the missing explanation.

## The tooltip

> Your account and its data live in the *&lt;region&gt;* region. To use an account in a different region, sign out and sign in at that region's URL.

Two things that are easy to blur, kept apart deliberately:

| User intent | Mechanism |
|---|---|
| "I have accounts in two regions, show me the other" | Sign out, sign in at that region's URL |
| "Move my existing account and its data there" | Operator-run transfer |

The copy promises only the first. Implying the UI can move an account's data between regions is the specific confusion worth avoiding — a region is where data lives, not a preference.

The wording is deliberately **deployment-neutral**: no hostnames, no Doubleword branding, no assumption of a particular topology. This dashboard ships to self-hosted installs as well as the gated console, and a single-region deployment reading it gets a statement that is vacuous rather than wrong.

## Implementation notes

`SidebarProvider` already wraps the app in a `TooltipProvider` (`ui/sidebar.tsx:129`), so Radix handles the `aria-describedby` wiring and no `sr-only` span is needed. Tooltip primitives come from the existing `components/ui` barrel rather than a second import path.

Unchanged: the indicator still renders only when a region is configured. A deployment serving one region has no use for the label, and an account's region is where its data lives, so a guessed default would be worse than nothing. That contract is now written down in the `Metadata::region` doc comment, since two dashboards depend on it.

## Why `dwctl/src/config.rs` is in this diff

release-please's root package is `dwctl`, and this is otherwise a dashboard-only change. The doc-comment edit gives the release a root-package change to attach to, so the dashboard build actually ships. It is a real comment improvement rather than a whitespace nudge — it records the hide-when-unset contract above.

## Related

- app-doubleword-private#200 — the same indicator in the public SPA
- doublewordai/internal#1589 — sets `metadata.region` on the US plane, the first deployment to report one

## Testing

`npm run build` and `npm run lint` clean in `dashboard/`; `cargo check -p dwctl` clean.

Refs COR-586


sejori:
Both Copilot findings were correct and are fixed in 24c2cc7.

**Keyboard reachability** — right, and it defeated the point of the change. Radix opens on focus as well as hover, but `TooltipTrigger asChild` over a plain `div` produces no tab stop, so the explanation was mouse-only. The trigger now carries `tabIndex={0}` and a `focus-visible` ring using the same `ring-ring` token the existing `tabs` component uses.

I kept the focusable `div` rather than switching to a `button`: the element does nothing on activation, so a real button would advertise an interaction that does not exist. A focusable element with `aria-describedby` is the more honest shape for static content that carries an explanation.

**Misleading comment** — also right. `TooltipProvider` governs behaviour like delay; the `aria-describedby` wiring comes from the trigger/content pair. The comment now says that, and states what the provider is actually load-bearing for here, which is that `SidebarProvider` already mounts one so this code did not have to.

Worth noting the sibling change in app-doubleword-private#200 does not share the first defect. That layout has no `TooltipProvider`, so it uses `title` plus an `sr-only` span — the text sits in the DOM in document order and needs no focus to be announced.

`npm run lint` and `npm run build` clean in `dashboard/`.

PR #1531, opened 2026-08-27
fix(dwctl): count pending requests outside the admission lock; index the batched demand branch

sejori:
## What

Two changes that together make it safe to re-enable `batches.pending_capacity_counts_enabled` in production (switched off in internal#1595 after batch submissions 500'd on the 60 s statement timeout).

1. **`fusillade-arsenal` migration `20260827000000`** — adds `idx_requests_active_batched_demand (model, service_tier, batch_id)` over the active, batched, templated, non-background working set, and drops the two `(batch_id, model)` partial indexes it supersedes (`idx_requests_active_sla_counts`, `idx_requests_active_non_priority_counts`). Net on production: +36 MB, −112 MB.
2. **`dwctl` `reserve_capacity`** — runs the pending-count query *before* taking the per-`(model, window)` advisory lock, on its own connection. To keep the documented swap-point race fail-safe with the reversed read order, it reads `since = now()` from the dwctl DB before the snapshot and `sum_active_by_model_window` now also counts reservations released at or after `since` — a batch that swaps mid-check is counted (at worst twice, never missed).

## Why

The count's cost is proportional to the active batched backlog of the requested models (4.9M rows for one model on 2026-08-25). Its batched branch heap-fetched every one of those rows because `service_tier` was not in any index. Worse, it ran inside the admission lock, so every concurrent submission for the same model queued behind it and shared its fate when it timed out.

## Measurements

Neon fork of the production DB (37M-row `requests`, 4.9M active batch-tier rows in 1,373 batches):

| Query | Before | With new index |
|---|---|---|
| admission shape (one model, 24h) | 6.5 s | 1.1–1.5 s |
| demand shape (all models, 24h) | 4.0 s warm / 14 s cold | 1.1–1.5 s |

Production `pg_stat_user_indexes` (cumulative): `idx_requests_active_sla_counts` 28K scans, last at 2026-08-25 13:33:55 — the last admission timeout — i.e. admission-only; `idx_requests_active_non_priority_counts` 2 scans ever. The three hot claim statements (`claim_batch_requests`, `claim_batchless_requests`, `claim_background_requests_by_kind`, from `pg_stat_statements`) were run with `EXPLAIN ANALYZE` on the fork before and after the swap: none used either dropped index, none adopted the new one, plans unchanged (confirmed by restoring the old indexes on the same statistics and re-running).

## Rollout

- Create the index `CONCURRENTLY` on production before deploying so the migration's `CREATE INDEX IF NOT EXISTS` is a no-op (statement is in the migration header). The drops are instant.
- After this ships: flip `pending_capacity_counts_enabled: true` in the internal production values (separate PR).

## Testing

- `cargo test -p dwctl -- capacity_reservations reserve_capacity_for_batch sla_capacity pending_counts_when`: 58 passed, including the new `test_reservations_released_since_are_counted`.
- `cargo test -p fusillade-arsenal background_schema_migration`: passes with the updated assertions.
- `cargo sqlx prepare --workspace` regenerated (two new entries), clippy clean on changed files, migration checksum fixture updated.


sejori:
Addressed the review — one commit with the two asks plus the fail-open, and fork verification for both empirical questions:

**1. Exact deployed variant, bound parameters.** Ran the query as a `PREPARE`d statement (all 11 params bound, `plan_cache_mode=force_custom_plan`) on a fresh fork of production with the migration state applied (`Include(vec![None])` renders as `$7='{}'`, `$8=true`, `$9='include'`): the planner chooses **Index Only Scan using `idx_requests_active_batched_demand`, Heap Fetches: 0**, for both the 24h and 1h windows. The predicate is proven not from the tier filter but from the query's own unconditional `r.service_tier IS DISTINCT FROM 'background'` clause, which matches the index predicate by exact clause equality — so provability does not depend on which tier filter the caller passes.

**2. Concurrent drops.** Agreed — the migration header now instructs `DROP INDEX CONCURRENTLY IF EXISTS` for both old indexes pre-deploy, alongside the `CREATE INDEX CONCURRENTLY`, leaving all three migration statements as no-ops on production.

**3. Fail-open.** Implemented: a failed or timed-out count logs an error and admits on active reservations only (same behavior as the flag being off, scoped to that one submission). Regression test drops `fusillade.requests` mid-test and asserts admission still succeeds.

**4. Taxonomy.** Confirmed on the fork: recent 1h-window batches carry `service_tier='flex'` on every request row (42,975 rows across 300 batches), 24h-window batches are all NULL-tier. So `Include(vec![None])` gates 1h admissions on reservations plus 24h-tier rows expiring inside the hour, not the existing 1h backlog — pre-existing behavior, now documented at the call site with the trade-off (widening to include `flex` would also count batchless Responses-API flex rows against batch capacity). Deferred to the counter-based estimator work.

On the burst caveat: agreed — un-serializing means N concurrent ~1.2s reads where the lock used to rate-limit to one. At current peaks (~20 creates/min) that's fine; a per-(model,window) count cache with a ~1s TTL is the natural next step if 50+ concurrent creates appear.

PR #1533, opened 2026-08-27
feat(inference): centralised parsing of out-of-band request params (COR-614)

jaedmunt:
Two things that I noticed/want to keep open:

1. Enabling this needs DWCTL_REQUEST_PARAMS__MODEL_SUFFIX=true added to the env of each node. It is off by default, so merging alone changes nothing. It must be set the same on every replica. If it is mixed, the same request is a flex request on one pod and a 404 on another which is a bit janky and while it feels safe, I don't really like that much.  I added this flag so we could roll it out but it will probably cause more overhead for making sure the env var is live on current new nodes, than if we have a staging or other check before it goes live system-wide.

2. Open question numba zwei (2). Rejections from this layer don't reach http_analytics so the 400 isn't in the customer request log. Pre-existing for all edge 4xx (401, 403, background contract), but inconsistent with the cache layer's cacheBreakpoint 400, which is currently logged. For now it emits a tracing::warn plus dwctl_http_requests_total{status}. Views welcome on where these belong as well plz.

Accepts service_tier and the caching marker via query param or model-name suffix, for clients that can only set base_url and model_name:

  POST /v1/chat/completions?serviceTier=flex
  `{ "model": "deepseek-ai/DeepSeek-V4-Flash-0731:flex" }`

It's parsed once in inference_middleware and normalised into the body, so no other layer changed and should work fine downstream. Exact alias match wins before any splitting, and a suffix is only stripped when what remains is a live alias, so nothing that routes today breaks.

No migration for the database should be needed. 

As a reminder this push leaves the url params part of the ticket out of scope and its just a way to sent a realtime-style request and bump it to flex tier.


jaedmunt:
Need to survey downstream compatibility with harnesses, whether they will accept `:flex` suffix or throw errors/strip it
This could also needs checks on whether flex-suffixed names are discoverable via the /list endpoint as usual

jaedmunt:
reverted to using `-dw-flex` instead of `:flex` to fit better with harnesses - such as avoiding conflict with harnesses parsing `:xhigh` and other parameters.

jaedmunt:
Closing the loop on the `/list` discoverability half of my earlier comment: it's already covered — `models_listing_is_unchanged_by_the_feature` (`dwctl/src/test/request_params.rs`) asserts `/ai/v1/models` returns only the base alias, with no `-dw-flex`/`-dw-priority` variants ever appearing in the listing. That's deliberate: a suffixed model resolves when a client types it, but the listing itself never grows a variant per model. Was in the original commit, just never replied to here.

PR #1557, opened 2026-09-02
feat(dwctl): real token-by-token streaming for flex requests via a Redis chunk relay

jaedmunt:
## What

Flex `stream:true` requests currently fake streaming: dwctl polls the row every 500ms, then replays the finished answer as one synthetic mega-delta. This relays the real per-token SSE chunks the daemon already sees (previously discarded) to the client-holding pod live, over Redis Streams.

Off by default (`flex_live_streaming.enabled: false`). When off, behavior is unchanged.

**Companion infra PR**: [doublewordai/control-layer-chart#101](https://github.com/doublewordai/control-layer-chart/pull/101) adds the chunk-relay Redis this needs in production. Linear: [COR-648](https://linear.app/doubleword/issue/COR-648).

## Why

- Real incrementality for clients doing incremental UI or early cancellation.
- Slowing the poll once live delivery is primary addresses the DB egress pattern behind COR-555, as a side effect of this being broader in scope than that fix alone.

## How

- `chunk_relay.rs`: Redis Streams relay. Publish side is non-blocking (called from a synchronous fold loop that can't `.await`). Subscribe side uses one shared reader task doing a batched `XREAD` across all active subscribers, instead of one blocking connection per client.
- `Sink::absorb` (`outbound_request.rs`) publishes each chunk additively, alongside the existing fold-into-one-body step, never instead of it.
- ZDR requests are included, not excluded: chunks are encrypted with the same per-request key the existing ZDR flow already manages.
- `flex_stream_response` races a relay task against the existing poll task; whichever delivers the terminal frame(s) first wins, and the other stands down immediately.

## Scope

Chat-completions surface only (Responses API not wired), first-attempt-success only (retry re-emission and mid-stream model escalation not handled), disconnect/cancellation unchanged. All deferred deliberately for this pass.

## Testing

Full end-to-end test: real daemon, real Redis, a real socket-bound mock upstream with genuine inter-chunk delays (not wiremock, it can't stream with real timing). Proves live delivery matches the mock's actual send time rather than the poll fallback, and a paired control test proves the flag-off path is byte-for-byte unchanged. Also added a test helper (`into_test_server_on_real_socket`) since no existing test previously exercised a genuine daemon-initiated loopback dispatch.

2210/2210 existing tests pass, lint clean.

## Not yet confirmed

Production Redis topology (single instance vs. cluster) and real throughput numbers on the actual deployed instance, flagged as follow-up before any wider rollout, not blocking for this prototype.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -158,6 +158,16 @@ jobs:
             sleep 1
           done
 
+      - name: Start Redis
+        # The flex live-relay tests (chunk_relay, outbound_request) need a
+        # real Redis; they default to redis://localhost:6379.
+        run: |
+          docker run -d --name redis -p 6379:6379 redis:7
+          for _ in $(seq 1 30); do
+            docker exec redis redis-cli ping && break
+            sleep 1
+          done
+
       - name: Install Rust
         uses: dtolnay/rust-toolchain@stable
         with:
--- a/config.yaml
+++ b/config.yaml
@@ -367,6 +367,21 @@ keystore:
  wrap_keys:
    dev-1: "local-dev-wrap-key-not-a-secret"
 
+# Flex live-streaming (prototype, default off): relays real per-token deltas
+# over Redis Streams instead of poll-and-replay. Absent/enabled:false leaves
+# today's flex streaming unchanged.
+#
+# Local dev: run `just redis-start`, then uncomment the block below.
+# flex_live_streaming:
+#   enabled: false
+#   chunk_relay:
+#     redis_url: "redis://localhost:6379"
+#     stream_ttl_secs: 600
+#     maxlen: 2000
+#     publish_channel_capacity: 1024
+#     reader_poll_interval_ms: 75
+#   poll_fallback_interval_ms: 5000
+
 # Batches API configuration
 batches:
   # Enable batches API endpoints (/files, /batches)
--- a/dwctl/Cargo.toml
+++ b/dwctl/Cargo.toml
@@ -108,10 +108,12 @@ hickory-resolver = { version = "0.26", default-features = false, features = [
   "tokio",
 ] }
 aes-gcm = "0.10.3"
-# Redis-backed keystore for zero-data-retention flex request keys.
+# Redis-backed keystore for zero-data-retention flex request keys, and the
+# flex live-streaming chunk relay (Redis Streams).
 deadpool-redis = "0.18"
 redis = { version = "0.27", default-features = false, features = [
   "tokio-rustls-comp",
+  "streams",
 ] }
 paste = "1.0"
 serde_with = "3.14.1"
--- a/dwctl/src/auth/middleware.rs
+++ b/dwctl/src/auth/middleware.rs
@@ -519,6 +519,7 @@ mod tests {
             response_store: state.response_store,
             image_normalizer: state.image_normalizer.clone(),
             keystore: state.keystore.clone(),
+            chunk_relay: state.chunk_relay.clone(),
         };
 
         let request = axum::http::Request::builder()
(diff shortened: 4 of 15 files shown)


jaedmunt:
In current state, this uses a local redis 

For production, this would become a default in the control-layer-chart, adding redis. 

 and a separate Elasticache set up in a very similar way to the ZDR one, for the AWS clusters. 
 
 Redis 8 would be better but Redis:7 could still be a sensible default

jaedmunt:
Local benchmark at 1k/5k/10k concurrent flex-streaming requests today, on top of the pipelining/sharding fix in this PR.

One tuning finding worth flagging, not proposing as a change: dropping `reader_poll_interval_ms` from the 75ms default to 10ms noticeably improved the live-relay-vs-poll-fallback race. Not recommending we actually change the shipped default off this though — the benchmark traffic is an artificially instant burst (not realistically paced), and going tighter still (2ms) made things worse via reader contention, so there's a real sensitivity here that could behave differently against real paced production traffic. Flagging for discussion if useful, not a suggested change.

| Concurrency | Errors | Mismatch | live_relay win rate |
|---|---|---|---|
| 1,000 | 0% | 0.04% | 99.96% |
| 5,000 | 6.1% | 0.8% | 99.19% |
| 10,000 | 13.6% | 0.01% | 100% |

Errors above are all client-side `ConnectTimeout` from this laptop's connection capacity (confirmed via zero Redis-side errors and zero rejected connections at every tier), not Redis or chunk_relay.

jaedmunt:
Extended to 20k/50k, kimi-k2.6:

| Concurrency | Errors | Mismatch | live_relay win rate |
|---|---|---|---|
| 1,000 | 0% | 0.04% | 99.96% |
| 5,000 | 6.1% | 0.8% | 99.19% |
| 10,000 | 13.6% | 0.01% | 100% |
| 20,000 | 82.2% | 0.18% | 99.83% |
| 50,000 | 88.7% | 0.15% | 99.82% |

Config: `reader_poll_interval_ms: 10` (local bench only), staggered client connection ramp. Errors above are client `ConnectTimeout` (this laptop), 0 Redis errors at every tier.

jaedmunt:
## Follow-up: real-model test on gotenks (Qwen3-4B via sglang)

Same fix, tested against genuine local inference (not a simulator) at 5000 concurrent. Included for comparison against the earlier gigatoken-paced results.

**Table 1 — gigatoken-paced simulator, shipped 75ms poll interval** (posted earlier)

| Concurrency | Errors | Mismatch | Near-collapse | live_relay win rate |
|---|---|---|---|---|
| 1,000  | 0%    | 0.19% | 0 | ~100%  |
| 5,000  | 0%    | 0.11% | 0 | ~100%  |
| 10,000 | 0%    | 0.00% | 0 | 99.8%  |
| 20,000 | 78.4%*| 0.00% | 0 | 99.96% |

**Table 2 — real Qwen3-4B (sglang, one RTX 4090), 5000 concurrent**

| Metric | Value |
|---|---|
| Successful completions | 4,229 (3.5%) |
| Connection failures | 116,145 (96.5%, `ConnectError`) |
| Real mismatches (>5 tokens off) | 8/4,229 (0.19%) |
| Near-total collapse | 0 |
| Redis rejected_connections | 0 |

Both tables' errors are client/connection-capacity limits, not Redis or chunk_relay — 0 rejected connections in every run. Table 2's 3/8 remaining mismatches look like a distinct, real interaction between the daemon's retry logic and chunk_relay under very long (34-52 min, uncapped) generations — noted for follow-up, not yet root-caused.

**Prod networking/hardware to check against this:**
- `net.core.somaxconn` on actual prod K8s nodes (128 locally, 4096 on gotenks — don't assume prod's value)
- dwctl pod's `ulimit -n` / container `nofile` limit
- Load balancer/ingress connection-concurrency settings (untested here — we hit dwctl directly)
- Postgres pool size × replica count vs. the managed DB's actual `max_connections`
- Whether HPA actually scales dwctl pods under load (we only tested one instance)

PR #1567, opened 2026-09-04
feat(fusillade): move overdue batchless graphs concurrently in the backfill

pjb157:
## Summary

Gives the batchless (retained-response) backfill mover real parallelism so the one-off drain of terminal batchless rows in `requests` runs in hours rather than months.

- New knob `batchless_archive_backfill_concurrency` (default 1) on `RetentionMaintenanceConfig`, the dwctl `DaemonConfig` (`DWCTL_BACKGROUND_SERVICES__BATCH_DAEMON__BATCHLESS_ARCHIVE_BACKFILL_CONCURRENCY`), the config-to-daemon wiring, and the startup controls snapshot. Validated `>= 1` while the backfill is enabled, in both dwctl and the daemon. The sweep worker stays sequential.
- The arsenal's overdue mover (`archive_overdue_batchless_responses`) now takes `concurrency`: it discovers one wave of distinct candidate graphs as before, then moves up to `concurrency` of them at the same time through a bounded `FuturesUnordered`. Each move is still its own transaction with the per-graph advisory lock, `FOR UPDATE SKIP LOCKED`, and read-back verification. The per-call byte budget is now an atomic reservation (`ByteBudget`), so concurrent moves cannot jointly exceed it and at most one oversized graph moves per pass, matching the sequential semantics. The graph budget is enforced by never dispatching more moves than could still fit.
- New gauge `fusillade_retained_response_movers_active{worker}`; `fusillade_retained_response_archive_may_have_more{worker}` still reflects the whole pass.
- README: rollout note plus a "one-off drain pod" recipe; `config.yaml` documents the knob.

### Why concurrency lives below discovery rather than as N daemon calls

Candidate discovery is an unlocked oldest-first read. N simultaneous passes would all discover the same head of the queue and then merely take turns on it through the advisory locks: safe, but no faster than one pass. One serial discovery followed by concurrent moves gives each mover its own graph. Concurrent passes (several pods, or a pod plus a drain pod) remain safe for the same reasons as before; they just do not add throughput.

### Throughput expectations

Wall clock per tick should shrink roughly with the concurrency until the primary's write path or the fusillade write pool saturates; discovery stays serial and is one index probe per candidate. Not measured here; measure on preview or staging before choosing a production value.

## Test plan

- [x] `cargo test -p fusillade` (daemon: backfill tick passes the configured concurrency to the overdue mover and reports the active-movers gauge, sweep never fans out, validation rejects 0 only while the backfill is enabled)
- [x] `cargo test -p fusillade-arsenal` (new `concurrent_overdue_passes_move_every_graph_exactly_once`: four concurrent passes with four movers each over twenty overdue singleton graphs; every graph moved exactly once, routes 1:1)
- [x] `cargo test -p dwctl`
- [x] `just lint rust`
- No `query!` macros changed, so no `sqlx prepare` needed. No migrations touched.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -690,6 +690,9 @@ background_services:
     # batchless_archive_backfill_enabled: false
     # batchless_archive_groups_per_tick: <positive bound>
     # batchless_archive_bytes_per_tick: <positive bound>
+    # batchless_archive_backfill_concurrency: 1  # Graphs the backfill moves at once per tick; the
+    #                                            # one-off drain lever (sweep stays sequential). Needs
+    #                                            # this many spare fusillade write connections.
     # retained_response_partitions_days_ahead: <positive runway>
     # Daily partition retirement irreversibly drops whole expired
     # retained-response partitions. It is independently disabled and needs a
--- a/dwctl/src/config.rs
+++ b/dwctl/src/config.rs
@@ -1844,6 +1844,15 @@ pub struct DaemonConfig {
     #[serde(default = "default_batchless_archive_bytes_per_tick")]
     pub batchless_archive_bytes_per_tick: i64,
 
+    /// Graphs the backfill worker moves concurrently within one tick. This
+    /// is the drain-throughput lever for the one-off historical drain: each
+    /// graph move is its own short transaction, so a tick's wall-clock time
+    /// is dominated by per-move round trips. The steady sweep is always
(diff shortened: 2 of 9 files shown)


pjb157:
Addressed the Copilot note: the active-movers gauge is now reset by a Drop guard (same pattern as the liveness gauge), so a panicked or cancelled tick can no longer leave it stuck. Also merged main.

PR #1574, opened 2026-09-04
feat(auth): make the active-organization cookie name configurable

sejori:
## Why

The org cookie's name was hard-coded as `dw_active_org` in three places: the read in `populate_org_context` (`dwctl/src/auth/current_user.rs`), the write in `set_active_organization` (`organizations.rs`), and the clear on logout (`auth.rs`). That is a problem for any deployment running several environments under one parent domain.

`cookie_domain` is commonly set to something like `.example.com` so that an app host and an API host can share org context. **A cookie scoped to a parent domain is offered by the browser to every host beneath it, and that is not configurable from the receiving side** — so a staging or second-region deployment sitting alongside production receives production's cookie as well as its own. Both arrive under the same name in one `Cookie` header, and the read here takes the first match:

    let org_id_str = parts.headers.get_all("cookie")
        ...
        .find_map(|cookie| cookie.strip_prefix("dw_active_org=")...)

Same-name cookies are ordered by path length and then **creation time** (RFC 6265 §5.4) — domain specificity is not a tiebreaker. So the winner is decided by which environment the user happened to sign in to first, not by which host the request actually reached. Worse, clearing the cookie on the narrower host cannot dislodge the parent-domain one, because `Set-Cookie` with `Max-Age=0` only clears the cookie matching the domain it is sent with.

Narrowing `cookie_domain` does not help: a parent-domain cookie always reaches its subdomains, so there is no domain arrangement that isolates environments sharing a parent. Only a distinct name gives each deployment its own slot.

## What

`auth.native.session.org_cookie_name`, defaulting to `dw_active_org`.

- Threaded through the read path as a new `populate_org_context` argument. The config read moves to the top of `from_request_parts` so the API-key branch (which runs before the previous `state.current_config()` call) can see it.
- Used by both write paths.
- Validated as a bare RFC 6265 token — the value is interpolated straight into a `Set-Cookie` header and matched as a `<name>=` prefix on read, so an empty string, whitespace, or a separator character would produce either a malformed header or a prefix that never matches.

Existing deployments are unaffected and need not set anything; the default is the name that was previously hard-coded.

## Tests

Two added in `config.rs`:

- `test_org_cookie_name_defaults_and_env_override` — the default is the historical name, and `DWCTL_AUTH__NATIVE__SESSION__ORG_COOKIE_NAME` overrides it.
- `test_invalid_org_cookie_name_rejected` — accepts the names we deploy (including a hyphenated per-preview one), rejects empty, whitespace, `;`, `=`, a newline, and a non-ASCII character.

Existing handler tests build request cookies with the literal `dw_active_org` against a default config, so they continue to exercise the default path unchanged.

## Deployment note

dwctl uses `deny_unknown_fields`, so this field crash-loops any image predating it. Deployments pinning `org_cookie_name` must be on a release carrying it. Anything that leaves it unset is decoupled from this release entirely.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -156,6 +156,20 @@ auth:
       cookie_name: "dwctl_session"
       cookie_secure: true
       cookie_same_site: "strict"
+      # Cookie carrying the caller's active organization.
+      #
+      # Give this a distinct value per environment whenever several
+      # deployments share a parent domain and cookie_domain is set. A cookie
+      # scoped to `.example.com` is offered to every host beneath it, so a
+      # staging or second-region deployment receives the neighbouring
+      # deployment's copy alongside its own — under the same name, in the same
+      # Cookie header. Same-name cookies are ordered by path length then
+      # creation time (RFC 6265 §5.4); domain specificity is not a tiebreaker,
+      # so the winner is whichever the user acquired first. Narrowing
+      # cookie_domain cannot prevent this: a parent-domain cookie always
+      # reaches its subdomains. Only a distinct name gives each deployment its
+      # own slot.
+      org_cookie_name: "dw_active_org"
     password_reset_token_duration: "30m"
 
   # Proxy header authentication
--- a/dwctl/src/api/handlers/auth.rs
+++ b/dwctl/src/api/handlers/auth.rs
@@ -374,8 +374,8 @@ pub async fn logout<P: PoolProvider>(State(state): State<AppState<P>>) -> Result
 
     // Also clear the active organization cookie
     let org_cookie = format!(
-        "dw_active_org=; Path=/; HttpOnly{}{}; SameSite={}; Max-Age=0",
-        secure, domain, session_config.cookie_same_site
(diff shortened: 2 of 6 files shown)


sejori:
Consumer side: doublewordai/internal#1912 — sets this per environment (`dw_active_org_staging`, `dw_active_org_us`, `dw_active_org_<namespace>` per preview) alongside the matching oauth2-proxy session-cookie rename. That PR must not merge until a release carrying this field is pinned in those environments.

sejori:
Fixed in 14490862 — thanks, the allocation was real and this is the per-request auth path.

Stripped the name and the `=` in two steps instead of formatting a joined prefix, and took the opportunity to borrow the name out of the config snapshot rather than `clone()` it, which was a second per-request allocation.

Worth noting the matching semantics are unchanged: the explicit `=` step still rejects a longer neighbouring name, so a `dw_active_org` reader does not match `dw_active_org_staging=…`. That exactness is the whole point of the field, so it would have been the thing to get wrong here.

PR #1698, opened 2026-09-07
fix(organizations): close the domain-claim hijack and the ownerless-workspace gap

sejori:
Supersedes #1658 (its commit is cherry-picked here onto current `main`; that branch was based on 11.4.3 and needs a rebase to merge).

## The reported bug (#1628)

Organization usernames are the domain-routing key. When the owner is on a personal email domain there is no domain to claim, and the fallback stored the user-supplied name verbatim into that same column — so a personal-email owner could type a third party's domain as the workspace name and `find_by_domain` would match it for every future signup at that domain.

The fix (Detail's, cherry-picked) makes the fallback an opaque `user~{suffix}`. The guard has to be that request-body text never reaches the username at all: `find_by_domain` matches on `username = $1 OR username LIKE $1 || '~%'`, so tightening only the bare-equality arm would leave `"acme.com~aaaaaaaa"` open. There's a regression test for each arm.

**No exploitation occurred.** Audited prod: every workspace owned by a personal-email account has a slugified name with no dot in it, and no workspace claims a domain its owner doesn't belong to.

## Naming (commit 2)

`username` is `{domain}~{suffix}` — or now `user~{suffix}` — and four responses rendered it as the workspace's *name*, so the organization switcher already showed `acme.com~a1b2c3d4`. Left alone, the opaque-username fix would have made that `user~a1b2c3d4` everywhere. All four now resolve `display_name` first, as the invite path already did. This also repairs the dashboard's `organization.name === inviteName` comparison, which was matching a routing key against a display name.

## Ownerless workspaces (commit 3)

Deleting a user scrubbed their row and dropped their keys but left their `user_organizations` rows behind. Deleting an organization's only owner left the workspace live, still claiming its domain, unadministrable — colleagues were pointed at it and could file join requests nobody could approve. `remove_member` has a last-owner guard for exactly this; account deletion walked past it.

- `Users::delete` hands each owned workspace to the longest-standing live member closest to the role (another owner → earliest admin → earliest ordinary member), or closes it the same way the account is closed when nobody is left: scrubbed, flagged deleted, keys hard-deleted, batches cancelled, fusillade data purged.
- `find_by_domain` skips workspaces with no live owner or admin, which contains the ones already stranded.

Prod has 119 of these. I checked before making routing skip them: none has ever had an API key, credit history or a single request, and only 8 have any live member at all. They're being cleaned up separately.

## Testing

Nine new tests: both hijack arms, the business-domain path still claiming its domain, the switcher naming, domain routing skipping an ownerless workspace, and the four succession outcomes. 205 `organizations::`/`users::`/`current_user` tests pass locally; `cargo fmt --check` clean.

## Follow-ups not in this PR

- `is_personal_email_domain` is a hardcoded ~35-entry list. Anyone on a freemail or consumer-ISP domain that isn't on it gets a legitimate claim on that whole provider, and `ORDER BY created_at ASC LIMIT 1` makes the first claimant permanent — the same "one workspace swallows every signup at a shared domain" outcome, through the sanctioned path. Wants a runtime-configurable list plus a data fix.
- Consider whether `users.username` should keep carrying routing, uniqueness and display at once; a dedicated `claimed_domain` column would remove the overload that caused this.


dwctl/src/db/handlers/users.rs
```diff
@@ -498,6 +498,88 @@ impl<'c> Repository for Users<'c> {
...
+                .fetch_optional(&mut *tx)
+                .await?;
+
+                match successor {
+                    Some(successor_id) => {
+                        sqlx::query!(
+                            r#"UPDATE user_organizations SET role = 'owner' WHERE organization_id = $1 AND user_id = $2 AND status = 'active'"#,
+                            org_id,
+                            successor_id
+                        )
```
sejori, line 548: All three applied in 43542fcc — the promotion UPDATE now filters on `status = 'active'`, the prod count is out of the test comment, and the wording is fixed. Thanks.

dwctl/src/auth/utils.rs
```diff
@@ -31,53 +31,440 @@ pub fn default_display_name(email: &str) -> String {
...
 pub fn email_domain(email: &str) -> Option<String> {
-    email.rsplit_once('@').map(|(_, domain)| domain.to_ascii_lowercase())
+    let (_, domain) = email.rsplit_once('@')?;
+    // A trailing dot is the DNS root and makes "acme.com." a different string
+    // to "acme.com", which would claim the same company twice.
+    let domain = domain.trim().trim_end_matches('.').to_ascii_lowercase();
+    is_claimable_domain_shape(&domain).then_some(domain)
+}
+
+/// Whether `domain` is shaped like a DNS name we are willing to route on.
```
sejori, line 41: All four applied in 18871376.

Both routing findings were correct and I'd underweighted the first one when I saw it myself earlier — I reasoned that `user` could never be a real domain and moved on, without checking that nothing validates the domain half. Fixed at the source: `email_domain` now requires two or more DNS labels of ASCII alphanumerics and inner hyphens, and `find_by_domain` re-checks rather than trusting call sites, since it's the function carrying the LIKE. Raw IDN is rejected too. Regression tests for `find_by_domain("user")`, `find_by_domain("%")` and the malformed-address shapes, as suggested.

Worth recording that the single-label path was about to become live rather than already being live: `user~{suffix}` doesn't exist in production yet, it arrives with this PR — and there are two production accounts on a single-label domain that would have matched it.

The key finding was a real hole and I've checked production: zero live org keys created by deleted users, so it was latent, never hit. Revoked in the same transaction now, matching `soft_delete_member_org_keys`.

And you were right that the ownerless test had stopped testing its own predicate — it now strands the workspace directly instead of going through `Users::delete`.

dwctl/src/db/handlers/users.rs
```diff
@@ -498,6 +498,142 @@ impl<'c> Repository for Users<'c> {
...
+                        // config-change NOTIFY as the delete trigger, so the proxy
+                        // still drops the key from its cache immediately - the
+                        // hard delete was buying nothing here that the soft one
+                        // does not.
+                        sqlx::query!(
+                            r#"
+                            UPDATE users
+                            SET email = $1, username = $2, display_name = NULL, avatar_url = NULL,
+                                is_deleted = true, updated_at = NOW()
+                            WHERE id = $3 AND user_type = 'organization' AND is_deleted = false
```
sejori, line 572: Pass 3 — all seven applied in ca032962. Two were real bugs and I've added a regression test for the first:

**Foreign key.** Confirmed: `connections.api_key_id` → `api_keys(id)` is NO ACTION in production. A workspace with a connection would have raised a violation on the hard delete and rolled back the entire account deletion, so the account could not be deleted at all — worse than the key surviving, which is what the hard delete was there to prevent. Now revoked in place. Worth noting `api_keys_notify_update` fires the same config-change NOTIFY as the delete trigger, so the proxy still drops it immediately; the hard delete wasn't buying anything the soft one doesn't. New test creates a connection pinning an org key and asserts the account still deletes.

**Collision.** Agreed, and the asymmetry is the point: the claimed arm's namespace is per-company so eight characters are ample, but the opaque arm puts every domainless workspace in one namespace against a UNIQUE column. Full UUID there, eight characters kept for the claimed arm.

**Pre-read and purge durability.** Both correct. The read is now all memberships rather than owned workspaces, so a mid-flight promotion can't escape it, and the post-delete check still decides what was actually closed. Cancellation is best-effort with the purge always enqueued — as you say, after the account row commits there is nothing left that could rediscover the workspace.

**Three doc claims.** All three were mine and all three were wrong: `display_name` genuinely can be null, and production has such rows, so "never the bare username" contradicted the fallback on the line above it. And the `find_by_domain` comment described stranding as current behaviour when this PR is what stops it. Rewritten to describe the legacy rows they actually cover.

PR #1699, opened 2026-09-07
fix(fusillade): bound retained response page scans

pjb157:
## Summary

Fix slow Responses pages for owners with large live and retained request histories. The retained arm could join broad routing tables and sort an owner's entire history before applying the page limit. A nullable-owner predicate also prevented owner-leading index conditions in generic prepared plans.

- Use distinct owner-scoped and unscoped SQL shapes, keeping all values bound.
- Validate retained partition metadata once in a materialized CTE.
- Read retained requests in chronological index order: validated retained requests are terminal, so their active-first rank is constant.
- Keep route validation and live-row deduplication as correlated indexed probes before each arm's limit.
- Preserve filtering, ordering, pagination, retention checks, and live-copy precedence. No schema migration or API change.

## Validation

- `cargo test -p fusillade-arsenal`: 371 passed, 2 ignored doctests.
- Rust formatting and Clippy passed.
- Full `just test rust` could not finish: the local disk filled while compiling the `dwctl` test binary.
- Full `just lint rust` reached SQLx verification, which reported nullable-metadata drift for an unchanged deployed-model query (`query-aabf664...`). Its cache has not been changed in this PR.
- Added a PostgreSQL `EXPLAIN ANALYZE EXECUTE` regression with forced generic plans and 4,000 retained rows. The original query visited 2,000 rows for a ten-row owner page; the rewritten query passes a fewer-than-100-row bound in both sort modes.
- Added page-by-page integration coverage for terminal-state timestamp ties, mixed live/retained ordering, owner-scoped and admin listings, and invalid routes preceding valid results.

## Scope

This fixes the page query independently of the existing count-budget fallback. Offset pagination still requires work proportional to the offset; selective residual filters and unscoped listings are not guaranteed constant-time. Cursor pagination is a separate follow-up.

### DIFF
--- a/fusillade-arsenal/src/postgres/retained_response.rs
+++ b/fusillade-arsenal/src/postgres/retained_response.rs
@@ -1268,7 +1268,20 @@ async fn count_requests_with_budget<P: PoolProvider>(
     Ok(total)
 }
 
-fn list_requests_page_sql(active_first: bool) -> String {
+fn list_requests_page_sql(active_first: bool, owner_scoped: bool) -> String {
+    // Separate SQL shapes let generic prepared plans use the owner-leading
+    // indexes. A nullable-owner OR cannot become an index condition in those
+    // plans. Values remain bound, including the NULL owner for admin listings.
+    let live_owner = if owner_scoped {
+        "request.created_by = $1::text"
+    } else {
+        "$1::text IS NULL"
+    };
+    let retained_owner = if owner_scoped {
+        "object.created_by = $1::text"
+    } else {
+        "$1::text IS NULL"
+    };
     let order_clause = if active_first {
         "CASE sort_state WHEN 'processing' THEN 0 WHEN 'claimed' THEN 1 WHEN 'pending' THEN 2 ELSE 3 END ASC, sort_created_at DESC, sort_id DESC"
     } else {
@@ -1283,14 +1296,34 @@ fn list_requests_page_sql(active_first: bool) -> String {
     } else {
         "request.created_at DESC, request.id DESC"
     };
-    let retained_order = if active_first {
-        "CASE object.state WHEN 'processing' THEN 0 WHEN 'claimed' THEN 1 WHEN 'pending' THEN 2 ELSE 3 END ASC, object.created_at DESC, object.object_id DESC"
-    } else {
-        "object.created_at DESC, object.object_id DESC"
-    };
+    // RetainedRequestPayloadV1::validate requires a terminal request. Every
+    // retained row therefore has the same active-first rank; omitting that
+    // constant key permits the existing owner/created_at/object_id index scan.
+    let retained_order = "object.created_at DESC, object.object_id DESC";
     format!(
         r#"
-        WITH candidates AS (
+        WITH valid_buckets AS MATERIALIZED (
+            SELECT bucket.delete_on
+            FROM retained_response_buckets bucket
+            JOIN pg_namespace namespace
+              ON namespace.nspname = bucket.partition_schema
+            JOIN pg_class child
+              ON child.relnamespace = namespace.oid
+             AND child.relname = bucket.partition_table
+             AND child.oid = bucket.partition_oid
+            JOIN pg_inherits inheritance
+              ON inheritance.inhrelid = child.oid
+             AND NOT inheritance.inhdetachpending
+            WHERE bucket.state = 'active'
+              AND bucket.partition_schema = current_schema()
+              AND bucket.partition_table =
+                  'retained_response_objects_d' || to_char(bucket.delete_on, 'YYYYMMDD')
+              AND inheritance.inhparent =
+                  to_regclass(format('%I.retained_response_objects', current_schema()))
+              AND pg_get_expr(child.relpartbound, child.oid) = format(
+                  'FOR VALUES FROM (%L) TO (%L)', bucket.delete_on, bucket.delete_on + 1
+              )
+        ), candidates AS (
             (SELECT
                 request.id AS sort_id,
                 request.state AS sort_state,
@@ -1327,7 +1360,7 @@ fn list_requests_page_sql(active_first: bool) -> String {
                 NULL::jsonb AS payload
             FROM requests request
             WHERE request.created_by IS NOT NULL
-              AND ($1::text IS NULL OR request.created_by = $1)
+              AND {live_owner}
               AND ($2::text IS NULL OR request.state = $2)
               AND ($3::text[] IS NULL OR request.model = ANY($3))
               AND ($4::timestamptz IS NULL OR request.created_at >= $4)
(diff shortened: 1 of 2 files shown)


fusillade-arsenal/src/postgres/retained_response.rs
```diff
@@ -1268,7 +1268,20 @@ async fn count_requests_with_budget<P: PoolProvider>(
...
+    // plans. Values remain bound, including the NULL owner for admin listings.
+    let live_owner = if owner_scoped {
+        "request.created_by = $1::text"
+    } else {
+        "$1::text IS NULL"
+    };
+    let retained_owner = if owner_scoped {
+        "object.created_by = $1::text"
+    } else {
+        "$1::text IS NULL"
```
sejori, line 1283: **The unscoped/admin path is not improved — it still times out.**

With `$1::text IS NULL`, the retained arm's index condition degrades to `created_by IS NOT NULL`, which yields no `created_at` ordering. So the planner falls back to `Append` + `Sort` over the entire retained history instead of `Merge Append`, and the LIMIT still can't be pushed down:

```
-> Limit
     -> Sort   Sort Key: object.created_at DESC, object.object_id DESC
          -> Result   One-Time Filter: ($1 IS NULL)
               -> Nested Loop Semi Join
                    -> Append          <-- not Merge Append
                         -> Index Scan using ..._created_by_created_at_object_idx3
                              Index Cond: (created_by IS NOT NULL)
```

Measured on prod: `ERROR: canceling statement due to statement timeout` at both a 60 s and a 240 s budget. So for a platform manager browsing `/responses` without an active organization, this page is unchanged.

I think the missing piece is an ordering index that doesn't lead with the owner, on the partitioned parent so `ensure_retained_response_partition` propagates it to new children:

```sql
CREATE INDEX idx_retained_response_objects_created
ON retained_response_objects (created_at DESC, object_id DESC)
WHERE object_kind = 'request';
```

That's a full prefix of the arm's `ORDER BY`, so it should give the same `Merge Append` the owner-scoped case now gets. Happy to raise it as a follow-up migration if you'd rather keep this PR to the query rewrite.

fusillade-arsenal/src/postgres/retained_response.rs
```diff
@@ -1327,7 +1360,7 @@ fn list_requests_page_sql(active_first: bool) -> String {
                 NULL::jsonb AS payload
             FROM requests request
             WHERE request.created_by IS NOT NULL
-              AND ($1::text IS NULL OR request.created_by = $1)
+              AND {live_owner}
```
sejori, line 1363: **With a date range set, the live arm becomes the whole cost.**

Same owner-scoped page as above but with `created_after` + `created_before` bound: 24 s first run, ~2.5 s warm on repeats. The breakdown is lopsided — of a 3,286 ms run, 3,279 ms is this arm:

```
-> Subquery Scan on "*SELECT* 1"   (actual time=3279.115..3279.136 rows=10)
     -> Limit
          -> Index Scan using idx_requests_user_active_sort on requests request
               Index Cond: (created_by = $1)
```

The retained arm in the same run is 4.7 ms, so this PR's work is doing its job; the residue is that `created_at` is a non-boundary column in `idx_requests_user_active_sort` (`created_by`, `CASE state`, `created_at DESC`, `id DESC`). With the owner bound it still walks every entry newer than `created_before` before reaching the range, which for a high-volume owner is a lot of index pages.

Consistent with "selective residual filters ... are not guaranteed constant-time" in Scope — just noting it's seconds rather than tens of milliseconds, and that it's now the dominant term whenever the date picker is used.

fusillade-arsenal/src/postgres/retained_response.rs
```diff
@@ -1283,14 +1296,34 @@ fn list_requests_page_sql(active_first: bool) -> String {
...
     };
-    let retained_order = if active_first {
-        "CASE object.state WHEN 'processing' THEN 0 WHEN 'claimed' THEN 1 WHEN 'pending' THEN 2 ELSE 3 END ASC, object.created_at DESC, object.object_id DESC"
-    } else {
-        "object.created_at DESC, object.object_id DESC"
-    };
+    // RetainedRequestPayloadV1::validate requires a terminal request. Every
+    // retained row therefore has the same active-first rank; omitting that
+    // constant key permits the existing owner/created_at/object_id index scan.
+    let retained_order = "object.created_at DESC, object.object_id DESC";
```
sejori, line 1302: The correctness of dropping the constant rank key rests on "every retained row is terminal", but nothing in the schema enforces that — the only `CHECK` on `retained_response_objects` is `object_kind_check`. `RetainedRequestPayloadV1::validate` guards the payload at write time, while `state` here is a separately denormalized column, so a row that ever diverged would silently mis-order rather than fail loudly.

Production is clean today (sampling one active partition: completed and failed only, no other states), so this is hardening rather than a bug. Would a constraint be worth adding, so the invariant is self-enforcing and the planner can see it too?

```sql
ALTER TABLE retained_response_objects
  ADD CONSTRAINT retained_response_objects_terminal_state_check
  CHECK (object_kind <> 'request' OR state IN ('completed', 'failed', 'canceled'));
```

fusillade-arsenal/src/postgres/retained_response.rs
```diff
@@ -1357,37 +1390,27 @@ fn list_requests_page_sql(active_first: bool) -> String {
...
-              AND bucket.partition_schema = current_schema()
-              AND bucket.partition_table =
-                  'retained_response_objects_d' || to_char(bucket.delete_on, 'YYYYMMDD')
-              AND inheritance.inhparent =
-                  to_regclass(format('%I.retained_response_objects', current_schema()))
-              AND pg_get_expr(child.relpartbound, child.oid) = format(
-                  'FOR VALUES FROM (%L) TO (%L)', bucket.delete_on, bucket.delete_on + 1
+            FROM retained_response_objects object
+            WHERE object.object_kind = 'request'
+              AND object.delete_on IN (SELECT delete_on FROM valid_buckets)
```
sejori, line 1395: Minor, and I don't think it needs changing now — worth knowing about.

Because `valid_buckets` is `MATERIALIZED`, its `delete_on` values aren't known at plan time, so this becomes a semi-join evaluated *above* the append rather than a partition-pruning quals. Every page therefore descends into every active bucket's partition:

```
-> Nested Loop Semi Join
     Join Filter: (valid_buckets.delete_on = object.delete_on)
     -> Merge Append
          -> Index Scan ... object_1
          ...
          -> Index Scan ... object_37
```

That's most of the 58 ms first-touch cost on the owner-scoped page, and it's cheap at the current bucket count. But it scales with the retention window rather than with page size, so if the number of active buckets grows substantially this is the term that grows with it. `MATERIALIZED` is doing real work here (it's what keeps the catalog validation out of the per-row path), so I'm not suggesting dropping it — just noting where the next ceiling is.

sejori (commented):
Verified this against production data rather than review-by-reading: I took `list_requests_page_sql` from this branch, ran the generated SQL through `PREPARE`/`EXECUTE` under `SET plan_cache_mode = force_generic_plan` (matching how sqlx executes it), and `EXPLAIN (ANALYZE, BUFFERS)`-ed each shape.

**The owner-scoped fix is real and large.** Default page, `active_first=true`, `service_tiers` filter, no date range, for the highest-volume owner in prod:

| | main | this branch |
|---|---|---|
| owner-scoped default page | **124.7 s** | **58 ms** first touch, 3-5 ms warm |

The retained arm plans exactly as intended — `Merge Append` over per-partition `(created_by, created_at DESC, object_id DESC)` index scans, LIMIT applied inside the arm, route/group validation and live-copy dedup as `SubPlan` index probes in the filter:

```
Limit
  -> Nested Loop Semi Join
       -> Merge Append   Sort Key: object.created_at DESC, object.object_id DESC
            -> Index Scan using ..._created_by_created_at_object_idx3
                 Index Cond: (created_by = $1)
                 Filter: ... EXISTS(SubPlan 2) AND (NOT EXISTS(SubPlan 4)) AND EXISTS(SubPlan 3)
```

The 4.75M-row `Seq Scan on retained_response_request_routes` that drove the old plan is gone. For context on what this replaces, `main` today serves that same page in ~124.6 s and the browser abandons it — nginx logs `499` against `console.doubleword.ai/responses`, tens of times a day for at least the last two weeks.

I also want to flag a bonus the description undersells: splitting the owner-scoped and unscoped shapes doesn't just fix generic plans, it drags the **live** arm off `idx_requests_active_first_tier` onto `idx_requests_user_active_sort` even in custom plans. Removing the nullable-owner `OR` did more than advertised.

Two measured gaps below, both of which the Scope section already concedes — I'm recording numbers so we can decide whether they're follow-ups or blockers, not asking for them here. Plus two smaller review notes.

Not blocking; approving in spirit.

sejori:
Raising the date-range cost from my inline note on the live arm (`AND {live_owner}`) as its own item, because I think it's fixable **here** rather than as a follow-up — it's the same function, and the fix is the same insight this PR already applies to the retained arm.

### The measurement

Owner-scoped page for the highest-volume owner in prod, `active_first=true`, `service_tiers` filter, under `force_generic_plan`:

| filters | this branch |
|---|---|
| no date range | 58 ms first touch, 3-5 ms warm |
| `created_after` + `created_before` | **24 s first run, ~2.5 s warm** |

Of a 3,286 ms warm run, 3,279 ms is the live arm and 4.7 ms is the retained arm. So this PR's work is doing its job; the residue is entirely on the live side.

### Why

`idx_requests_user_active_sort` is `(created_by, CASE state ..., created_at DESC, id DESC, service_tier)`. With `created_by` bound, the `CASE` column is an *ordering* key but not a *bounded* one, so `created_at` is the third column with an unbounded column above it. On PostgreSQL 17 that can't become a boundary condition (no btree skip scan until 18), so it degrades to an index filter: the scan still walks every entry newer than `created_before` before reaching the requested range. For a high-volume owner that's tens of thousands of index pages, and on Neon each miss is a pageserver round trip.

Same query with `active_first=false` uses `idx_requests_user_created_sort` and returns in **3.8 ms**, which isolates the cause precisely — it's the rank key in the sort, not the date predicate.

### Why I don't think a new index is the answer

The index that would fix it already exists (`idx_requests_user_created_sort`); the planner just can't use it while the rank key sits in the `ORDER BY`. Adding another permutation would mean a third ~4 GB index on `requests` to serve a rank key that is currently doing almost nothing:

```
SELECT state, count(*) FROM requests
WHERE created_by IS NOT NULL AND state IN ('pending','claimed','processing');
-- processing | 1
```

One row. The entire non-terminal batchless population.

### Proposed fix — the retained arm's own argument, applied to the live arm

This PR already says, for the retained side:

> validated retained requests are terminal, so their active-first rank is constant

The same is true of *terminal live rows*. So the live arm can split the way the union already does:

```sql
(SELECT ... FROM requests request
 WHERE <filters> AND request.state IN ('processing','claimed','pending')
 ORDER BY CASE request.state ... END ASC, request.created_at DESC, request.id DESC
 LIMIT $7 + $8)
UNION ALL
(SELECT ... FROM requests request
 WHERE <filters> AND request.state NOT IN ('processing','claimed','pending')
 ORDER BY request.created_at DESC, request.id DESC
 LIMIT $7 + $8)
```

The second arm now has a constant rank, so it orders by `created_at DESC, id DESC` only and `idx_requests_user_created_sort` serves it with `created_at` as a genuine boundary column. The first arm is bounded by the partial state indexes over a population of single digits. The outer `ORDER BY` merges them exactly as it already merges live against retained — and the correctness argument is the one already made in this PR, not a new one.

No migration, no new index, and it collapses the `active_first=true` and `active_first=false` paths onto the same access pattern.

I'm going to trial this on a Neon branch against a prod-sized copy and post the `EXPLAIN ANALYZE` numbers. Happy for it to land here or as a follow-up — flagging it now mainly so it's acknowledged rather than lost, since it survives this merge and is the dominant cost the moment anyone touches the date picker.

PR #1712, opened 2026-09-08
feat(fusillade): add configurable batch leak intervals

fergusfinn:
Small batches for models without live local capacity currently wait until the deadline ramp, even when an external provider could serve them. Add an optional one-request-per-bucket trickle so they can progress earlier. Full claims take priority over trickle rows at the batch-selection, model-capacity and total-claim limits.

Expose `background_services.batch_daemon.leak_enabled` (default true), `leak_interval_seconds` (default 60) and exact-model `model_leak_interval_seconds` overrides. These controls apply only to batches. Async/flex retains the existing `leaks_per_window` policy: 24 minutes for default-tier rows and 60 seconds for flex with the default setting. Disabling batch leaking restores the live gate while retaining the deadline escape hatch, including after restart.

Library callers opt into batch leaking with `with_leak_config`; the existing `claim_batch_requests` method and custom-backend fallback retain their non-leaking behavior. NULL and explicit `24h` windows share a bucket. Cooldowns remain process-local and scoped per user/model/window, with one immediate claim and no accumulated burst. The leak counter has `daemon` and `model` labels.

Validation:
- SQL regressions cover cooldowns, normalized bucket keys, legacy callers, live/unmanaged bypass and deadline priority at all three limits.
- Daemon tests cover unchanged async refill rates, model overrides, eventual batch refills, and disabled/unconfigured batch leaking across fresh daemon instances.
- Config tests cover defaults, YAML/environment overrides, the off switch and invalid intervals.
- Local PostgreSQL 17 `EXPLAIN ANALYZE` with 5,000 candidate batches and 5,000 cooldown entries: custom and generic plans scanned the cooldown set once (49 ms and 43 ms total respectively). This is a synthetic plan check; production data was not queried.

Checks passed: `just test rust` (3,276 passed, 31 ignored), `just lint rust -- -D warnings`, and `just lint ts`. Rust tests used an isolated local PostgreSQL database.


fusillade-arsenal/src/postgres.rs
```diff
@@ -2644,11 +2671,20 @@ impl<P: PoolProvider> Storage for PostgresRequestManager<P> {
                           AND (r.not_before IS NULL OR r.not_before <= $3)
                     )
                 ) ranked
+                WHERE claim_full OR leak_rank = 1
+            ),
+            selected_batches AS (
+                SELECT * FROM (
+                    SELECT rb.*, row_number() OVER (
+                        PARTITION BY model ORDER BY pr ASC, expires_at ASC, batch_id ASC
```
sejori: **Blocking: trickle batches can crowd out the deadline escape hatch.**

Leak batches and full-claim batches now compete in one `batch_rank <= $6` pool per model (`batch_claim_batch_size`, default 4), ordered by `pr`. With dwctl's `urgency_weight = 0.5`:

- idle user's leak batch, 20h to deadline -> `pr ~= 0.5*0 + 0.5*(72000/86400) = 0.417`
- busiest user's batch *inside its ramp* -> `pr ~= 0.5*1.0 + 0.5*0.003 = 0.50`

So four idle-user trickle buckets on a not-live model evict the one batch the ramp exists to rescue. Before this PR the pool only ever held full-claim batches, so this is a new regression — and it defeats the specific guarantee the escape hatch provides.

Cheap fix: `ORDER BY claim_full DESC, pr ASC, expires_at ASC, batch_id ASC` in this window, or give leak batches a slot budget separate from `batch_claim_batch_size`.

sejori (commented):
Reviewed against the original design (model_filters liveness gate + per-`(user, window_class, model)` leaky bucket + the `window_minutes ^ 0.56` deadline ramp). The SQL shape is right and the bucket-grain test coverage is genuinely good, but I don't think this is mergeable as-is: two bugs and two decisions, detailed inline.

**Blocking**
1. Trickle batches can crowd the deadline escape hatch out of the `batch_claim_batch_size` pool (postgres.rs `selected_batches`).
2. The leak bucket key is inconsistent between the `PARTITION BY` and the cooldown predicate for NULL `completion_window`, so one bucket can leak twice per interval.

**Needs a decision, not just a review**
3. This silently makes the *existing async* trickle 24x faster for default-tier rows — it isn't only a batch change, and the PR body doesn't mention it.
4. There's no off switch, and one knob drives both loops. Given the admission gate had to be flipped off in prod within days of shipping, I'd want an `enabled` flag or separate batch/async intervals.

**What's good:** `leak_rank` computed after the `WHERE` so cooldown filtering happens first; `LIMIT CASE WHEN sb.claim_full THEN sb.capacity ELSE 1 END` is the right one-row pull; `leaked` stamped only for non-full rows so ramp claims correctly consume no token; batch and async bucket keys don't collide (`'24h'`/`'1h'` vs `'default'`/`'flex'`); the additive builder keeps downstream struct literals compiling. The per-bucket SQL test (separate users/models/windows, live/unmanaged bypass, ramp bypass) covers the cases that actually matter.

fusillade-arsenal/src/postgres.rs
```diff
@@ -2578,17 +2609,18 @@ impl<P: PoolProvider> Storage for PostgresRequestManager<P> {
...
                 FROM (
                     SELECT g.model, g.capacity, b.id AS batch_id,
                            b.expires_at, b.created_at, b.created_by,
                            COALESCE(b.completion_window, '24h') AS window_class,
-                           calc.pr,
+                           calc.pr, gate.claim_full,
                            row_number() OVER (
-                               PARTITION BY g.model
+                               PARTITION BY g.model, COALESCE(b.created_by, ''),
+                                            b.completion_window, gate.claim_full
```
sejori: **Blocking: bucket key is inconsistent between the partition and the cooldown.**

This partitions on the raw, nullable `b.completion_window`, but the cooldown predicate (L2657) and `window_class` (L2617) both use `COALESCE(b.completion_window, '24h')`.

A NULL-window batch and an explicit `'24h'` batch for the same user+model therefore land in *different* partitions, both get `leak_rank = 1`, and both stamp the same `('24h')` bucket — two leaks per interval instead of one.

`PARTITION BY g.model, COALESCE(b.created_by, ''), COALESCE(b.completion_window, '24h'), gate.claim_full`.

dwctl/src/lib.rs
```diff
@@ -3720,7 +3720,8 @@ impl Application {
         let request_manager = Arc::new(request_manager);
         let postgres_daemon = Arc::new(
             fusillade::PostgresDaemon::from_store(request_manager.clone(), fusillade_daemon_config.clone())
-                .with_retention_maintenance(retention_maintenance_config),
+                .with_retention_maintenance(retention_maintenance_config)
+                .with_leak_config(config.background_services.batch_daemon.leak.clone()),
```
sejori, line 3724: **Needs a decision: this silently makes the existing async trickle 24x faster.**

`LeakConfig` is installed unconditionally here, and `stamp_leaks` applies it to *both* foreground loops, not just the batch one.

Batchless rows with no `service_tier` get `window_class = 'default'` -> `W = 24h`, so today's interval is `86400 / leaks_per_window = 86400/60 = 24 min`. After this PR it's 60s. Flex (1h) is unchanged.

That's a 24x increase in overflow to fallback providers on the main async path — likely the single biggest cost effect of this PR, and the body reads as a batch-only change. Either scope the fixed interval to the batch loop, or make the async change explicit with an estimate of the extra external spend.

config.yaml
```diff
@@ -643,9 +643,13 @@ background_services:
     # batch_claim_size: 0            # Max rows per batch-claim iteration; 0 = inherit claim_batch_size
     # batch_claim_batch_size: 4      # Max batches selected per model per iteration (spill-over pool)
     # batch_claim_interval_ms: 0     # Batch-claim loop cadence; 0 = inherit claim_interval_ms
-    # batch_claim_require_live: false # Require an explicit `live` model_filters event to batch-claim;
+    # batch_claim_require_live: false # Require `live` for full-rate batch claims outside the deadline ramp;
     #                                 # false treats no-event models (external providers) as live
-    # claim_ramp_exponent: 0.56      # Deadline-ramp curve: not-live batches become claimable within
+    # Batch and async requests on unavailable models each get one immediate claim,
+    # then one per interval, per user/model/window. Positive whole seconds only.
+    leak_interval_seconds: 60
```
sejori, line 653: **Needs a decision: no off switch, and one knob drives both loops.**

If batch leaking misbehaves in prod, the only lever is a very large `leak_interval_seconds` — which also disables the async trickle, and every bucket still gets a free immediate claim on each daemon restart or rollout (the cooldown is process-local).

The batch admission gate had to be flipped off in prod within days of shipping for exactly this reason. I'd want either an `enabled: bool`, or separate batch/async intervals so one can be backed out without the other.

Also: this lives under `background_services.batch_daemon.*` but governs async claims too. That name will mislead whoever touches it during an incident.

fusillade-arsenal/src/postgres.rs
```diff
@@ -2494,6 +2494,26 @@ impl<P: PoolProvider> Storage for PostgresRequestManager<P> {
         daemon_id: DaemonId,
         available_capacity: &std::collections::HashMap<String, usize>,
         user_active_counts: &std::collections::HashMap<String, usize>,
+    ) -> Result<Vec<Request<Claimed>>> {
+        self.claim_batch_requests_with_cooldown(
```
sejori: **`claim_batch_requests` quietly changed meaning for external callers.**

It now delegates to the cooldown variant with an *empty* set, so any caller of the old method gets one leaked row per bucket per call with **no throttle at all** — where previously it got the strict live-gate and no leak whatsoever.

That's a behavioural break on a published-crate method, and it fails open (more spend, not less). At minimum it needs a changelog callout; better would be to leave `claim_batch_requests` non-leaking and make the trickle strictly opt-in via the new method.

Same class of change: `batch_claim_require_live: true` goes from "exclude unmanaged models entirely" to "trickle them". The README covers it, but it's a semantic change to an existing flag.

fusillade-core/src/manager.rs
```diff
@@ -1390,6 +1380,31 @@ pub trait Storage: Send + Sync {
...
+    async fn claim_batch_requests_with_cooldown(
+        &self,
+        limit: usize,
+        batch_limit: usize,
+        daemon_id: DaemonId,
+        available_capacity: &std::collections::HashMap<String, usize>,
+        user_active_counts: &std::collections::HashMap<String, usize>,
+        leak_cooldown: &std::collections::HashSet<(String, String, String)>,
+    ) -> Result<Vec<Request<Claimed>>> {
+        let _ = leak_cooldown;
```
sejori, line 1399: The trait default silently discards `leak_cooldown`, so any custom backend that doesn't override this gets an unthrottled trickle rather than an error or the old behaviour. Combined with the `claim_batch_requests` delegation on the Postgres side, both compatibility paths fail open. Worth at least a doc line saying so explicitly — "preserves compatibility" reads as "preserves behaviour", which it doesn't.

fusillade/src/daemon/mod.rs
```diff
@@ -1975,8 +1984,9 @@ where
         let mut leaked_count = 0u64;
         for request in claimed {
             if let Some(stamp) = &request.state.leak {
-                let interval = std::time::Duration::from_secs_f64(
-                    (stamp.window_secs / leaks_per_window).max(0.0),
+                let interval = self.leak_config.as_ref().map_or_else(
+                    || Duration::from_secs_f64((stamp.window_secs / leaks_per_window).max(0.0)),
+                    |config| config.interval(&request.data.model),
```
sejori: `fusillade_leaky_bucket_leaks_total` is a single unlabelled counter that is now incremented by two loops running different policies against different bucket namespaces.

Please add a `daemon` label (`request_daemon` / `batch_daemon`) and ideally `model`. There's no global overflow budget backstopping this, so during a fallback-spend incident "which loop and which model is leaking" is exactly the cut you'll need, and right now it isn't recoverable from the metric.

fusillade-arsenal/src/postgres.rs
```diff
@@ -2616,20 +2643,20 @@ impl<P: PoolProvider> Storage for PostgresRequestManager<P> {
...
+                            COALESCE(mf.state = 'live', NOT $10::BOOLEAN)
                             OR (EXTRACT(EPOCH FROM (b.expires_at - $3))
-                                    <= power(GREATEST(EXTRACT(EPOCH FROM (b.expires_at - b.created_at)), 0.0) / 60.0,
-                                             $11::DOUBLE PRECISION) * 60.0)
-                          )
+                                <= power(GREATEST(EXTRACT(EPOCH FROM (b.expires_at - b.created_at)), 0.0) / 60.0,
+                                         $11::DOUBLE PRECISION) * 60.0)
+                        ) AS claim_full
+                    ) gate
+                    WHERE (gate.claim_full OR NOT EXISTS (
```
sejori: Minor / worth an EXPLAIN before rollout: this is `O(candidate batches x cooldown entries)` against an unindexed `unnest`. It's the same pattern as the batchless arm, but the batch candidate set is larger (batches-with-pending-work per model, across all capacity-eligible models) and the cooldown set grows with active buckets.

Given this query's history of 60s statement timeouts in prod, I'd want the plan checked on prod-shaped data rather than assuming the planner hashes it.

fusillade/tests/integration.rs
```diff
@@ -44,6 +44,133 @@ async fn mark_models_live_for_test(manager: &PostgresStore<TestDbPools>, models:
...
+            break;
+        }
+        assert!(
+            tokio::time::Instant::now() < deadline,
+            "tiny workloads failed to progress: {calls} calls"
+        );
+        tokio::time::sleep(Duration::from_millis(10)).await;
+    }
+    assert!(
+        started.elapsed() >= Duration::from_secs(2),
```
sejori: Minor: this test is wall-clock dependent — real sleeps plus `calls <= 3` while `elapsed < 1s`, `elapsed >= 2s`, and a 10s deadline. On loaded CI that's a flake waiting to happen, and fusillade already carries one load-sensitive integration test. The SQL-level tests already prove the gate; consider dropping the timing assertions here to just "it eventually refills".

PR #1747, opened 2026-09-10
fix(onwards): scrub upstream response headers that leak provider topology

sejori:
## What

Add `scrub_response_headers` to the onwards proxy — the response-side counterpart to `filter_headers_for_upstream`. Called immediately after the upstream response is received in `target_message_handler`, it strips:

- **Hop-by-hop headers** (RFC 7230 §6.1): `connection`, `keep-alive`, `proxy-authenticate`, `proxy-authorization`, `te`, `trailer`, `upgrade`.
- **The provider/edge fingerprint**: `set-cookie`, `x-generation-id`, `server`, `x-powered-by`, `via`, `alt-svc`, `nel`, `report-to`, `reporting-endpoints`, and the whole `cf-*` family by prefix.

## Why

Upstream response headers were passing through to API clients verbatim. A Cloudflare-fronted provider announces itself on every response through `cf-ray`, `cf-cache-status` and `server: cloudflare`, sets `set-cookie: __cf_bm=…`, and some upstreams append `x-generation-id` — all of it naming the third-party routing layer, and the cookie planting a provider session on our caller.

The request-side filter already strips hop-by-hop and auth headers from outbound requests; the response side was never scrubbed.

`alt-svc` is not only a leak. The upstream advertises *its own* HTTP/3 endpoints, and a client receiving that header on our response applies the advice to the onwards host instead.

### What is deliberately kept

- **`retry-after` and `x-ratelimit-*`** describe the upstream too, but a client cannot back off correctly without them.
- **`transfer-encoding`** is hop-by-hop per the spec but excluded from `HOP_BY_HOP`: outbound framing is hyper's to decide, and the body-rewriting paths already drop it where they set an explicit `content-length`. Documented on the const, and asserted in `test_scrub_response_headers_removes_hop_by_hop`.

### Prefix matching

`cf-*` is matched by prefix rather than enumerated — Cloudflare's family grows without notice, and this is the same treatment `filter_headers_for_upstream` already gives `sec-ch-ua*`.

## Coverage

Both the strict and non-strict paths flow through `target_message_handler`, so a single scrub call covers both. It sits before every branch that rebuilds the response from its parts (strict SSE peek, unary embedded-error scan, non-strict `SseBufferedStream`), so all of them inherit it. The existing custom-response-headers injection (`target.response_headers`) happens ~500 lines *after* the scrub, so operator-configured headers are unaffected — and an operator setting a scrubbed name explicitly still wins.

The RFC 7230 hop-by-hop set now lives in one `HOP_BY_HOP` const shared by both filters. It was duplicated verbatim; silent drift between a request-side and a response-side filter is the exact bug class this PR exists to fix.

## Validation

- `cargo check -p onwards --all-targets` — compiles clean
- `cargo clippy -p onwards --all-targets` on `@stable` — unchanged at 3 pre-existing warnings (`main.rs:24`, `strict/handlers.rs:87`, `handlers.rs:1598`), none in new code
- `cargo fmt --check` — clean
- `cargo test -p onwards` — 452 + 4 + 10 passing, 0 failed

### Tests

8 unit tests on the helper, including the two non-obvious properties: `HeaderMap::remove` drops *every* `set-cookie` value rather than the first (upstreams routinely send several), and the `cf-` prefix rule covers a header we have never seen.

4 end-to-end tests in `onwards::tests::response_header_scrubbing` drive a mock upstream that returns the full leaked header set, covering the unary path, the strict-mode buffer-and-rebuild, the streaming path (which also exercises the mock's own `connection: keep-alive`), and the operator-`response_headers` ordering. The scrub is a single call site, and the regression that matters is a future refactor moving or losing it — which the handler unit tests cannot see.

## Follow-up (not in this PR)

The strip list is a hardcoded constant. A self-hoster proxying a cookie-bearing upstream has no escape hatch, and `x-generation-id` is an OpenRouter-specific name baked into a general-purpose crate. Making it per-target configurable alongside the existing `response_headers` would need plumbing through the four `Target`/`Pool` structs in `target.rs` and the builders — deliberately left out to keep this diff reviewable.


sejori:
Superseded by [doublewordai/sso-stack#420](https://github.com/doublewordai/sso-stack/pull/420), which strips the same headers at the edge proxy.

That is the right choke point for our deployments: `main` fronts console + api and `public` fronts app, so every customer-facing response passes through one of them, and a single change covers any backend we put behind them — current or future — without an onwards release.

#420 now carries the full header list from this PR (`set-cookie`, `x-generation-id`, `server`, `x-powered-by`, `via`, `alt-svc`, `nel`, `report-to`, `reporting-endpoints`, and the `cf-*` family), so closing this loses no coverage at the edge.

Two things it does **not** carry, recorded here so they are not lost:

1. **No prefix matching.** nginx's `proxy_hide_header` takes exact field names, so the `cf-*` family is enumerated in #420 rather than matched by prefix. A `cf-` header Cloudflare adds later is covered here but not there. Worth a periodic look at the list.

2. **Only traffic through that nginx is covered.** onwards is also used as a library (`onwards-fusillade`, batch dispatch) and as a standalone crate by self-hosters who do not run sso-stack. Those paths keep the upstream headers. This is a non-issue for our own async paths — response headers are not persisted with batch output — but it does mean the OSS crate still forwards a provider's `Set-Cookie` verbatim.

The hop-by-hop half of this PR is genuinely redundant: nginx strips RFC 7230 hop-by-hop headers natively.

If we ever want the prefix rule or want the crate correct standalone, the branch is `fix/scrub-upstream-response-headers` and it was green (23/23) at 3a0488ed.

PR #1790, opened 2026-09-16
feat(onwards): measure first-token latency per provider role

sejori:
Adds observation metrics for the existing first-token failover path, providing evidence for the proposed load-aware controller without changing provider selection.

- Record observed strict-mode SSE first-frame latency as `onwards_first_token_seconds`, labelled by model, resolved pool and preferred/alternate provider role.
- Count failover-deadline expiries in `onwards_first_token_breaches_total`, including waits for headers. These censored observations never enter the histogram; independent request timeouts and upstream failures do not count as breaches.
- Keep `[DONE]`-only streams valid without treating the sentinel as a latency sample.
- Document partial sampling coverage, the distinction between failover deadlines and controller budgets, and the need for fresh observations after share changes.
- Document the embedded metrics in dwctl so the application release includes the observation change.

The design page remains a plan for the controller. Per-model configuration, complete controller observation coverage, and share-based selection are not implemented here.

Validation: onwards unit, integration and documentation tests, including nine new end-to-end metric regressions; workspace Rust tests and lint; TypeScript lint. Sampling tests cover sentinel-only streams, keep-alives, header/body timeouts, errors, named pools, and non-strict/peek-limit exclusions.


sejori:
All six points were valid and are addressed in 4ab1631a3. Three I confirmed against the code before acting on them.

**1. Sampling paths are not equivalent (strict-mode gating)** — Confirmed, and the most serious of the six. The lead-frame branch is gated on `(200..300).contains(&status) && state.targets.strict_mode`, so non-strict SSE is forwarded with no lead-frame peek and would yield no uncensored samples at all. The page now says the first-token histogram is populated only for strict-mode SSE, states that header arrival is *not* a first-token sample, and records that covering non-strict traffic needs a pass-through-safe observer which must not re-frame or buffer — carried as an open question rather than assumed free, since the pass-through path avoids that buffering deliberately.

**2. `weighted_random` is not a proportional selector** — Confirmed and corrected. `select_excluding` dispatches `WeightedRandom` to `select_least_connections`, which ranks by lowest `active/weight` and consults weights only to break ties, so weights are a normaliser rather than a splitter. The weight-folding idea is gone. Selection now seeds the exclusion set with the preferred provider with probability `1 - f` before the first attempt: uniform across both strategies, and it leaves the attempt budget, exclusion mechanics and cascade restart untouched. Preferred provider is now defined as first in definition order, and per-provider shares in pools larger than two are explicitly out of scope.

**3. Breach signal undefined; a timeout is censored** — Agreed, and the sharpest point. A deadline expiry establishes only that first-token latency exceeded the deadline, so recording the deadline value as a latency would bias every statistic drawn from the histogram. The page now separates uncensored samples from breaches, exports them as distinct metrics, and makes increase require in-budget uncensored samples while decrease is driven by breaches and slow samples. Implementation step 2 and the testing section are now consistent with each other.

**4. `[DONE]` can register as a first token** — Confirmed. `classify_sse_event` maps a `[DONE]` payload to `Data`, and `read_lead_frames` sets `saw_data` on any `Data` frame, so a zero-token stream could count as success. A sample now requires a data frame that is not `[DONE]`, with a test for it.

**5. Metric label cardinality** — Confirmed, and the constraint is firmer than described: the recorder carries an explicit instruction not to enable idle-timeout or eviction, because the autoscaler reads an absent in-flight series as genuinely zero. Series therefore persist for the process lifetime, so provider URLs are unusable as labels and have been removed. Metrics are alias-labelled, with a bounded preferred/alternate role label if per-provider detail is ever needed.

**6. Low settled share as capacity evidence** — Agreed, overclaimed. Downgraded to an indicator, conditional on classifying hard failures separately: connection errors, non-2xx responses and embedded upstream errors must not decay the share, otherwise an outage reads as a capacity limit. Added a test for that classification.

Still design-only — no implementation in this PR.

sejori:
Second review actioned in a5f6e0db9. All five findings were valid; the first one undercut the design's central claim.

**1. A single breach cannot drive convergence** — Agreed, and this was the important one. Any realistic first-token distribution has a tail, so at *every* sustainable share some requests exceed the budget. Decreasing on each breach would decay the share to its floor regardless of real capacity, and would decay it faster at higher request volumes, making the control a function of traffic rather than of service quality. The controller now targets a **breach rate** against the budget, measured over a sliding window with a minimum sample count, and converges on the largest share whose breach rate still meets that target. That is a service level it can actually hold.

**2. The bias was not uniform across strategies, and exclusion-seeding was unsafe** — Confirmed against the code, including the qualifier. `SelectIter::next` clears exclusions only when `select_excluding` returns `None`, so while any alternate remains selectable a seeded exclusion persists and the preferred provider is unreachable for the rest of that request. Separately, under `WeightedRandom` leaving a provider eligible does not make it first, because `select_least_connections` ranks by `active/weight` and uses weights only to break ties — so the realised preferred-first rate would sit below `f` by a load-dependent margin. The design now uses a **first-attempt-only override** rather than mutating the exclusion set, and is **scoped to `Priority`**, with `WeightedRandom` deferred and the reason recorded as an open question.

**3. Control input must be scoped to the preferred provider** — Agreed. Once `1 - f` of traffic goes to alternates, letting their observations move the share means a slow alternate demotes a healthy preferred provider and a fast one ramps up a struggling one. Observations are now attributed to the provider actually attempted, and only preferred-provider attempts adjust the share. Alternate observations are still exported, since they are useful for comparing providers, but they are inert as control input. This is also why the share floor matters: it is what keeps preferred-provider observations flowing at a low share.

**4. State adoption needs an identity check** — Agreed. `adopt_active_counter` is safe precisely because it matches counters per provider identity; a single pool-level share has no such matching, so blind adoption would apply a share learned about one upstream to whatever now sits first after a reorder or replacement. Adoption is now conditional on the preferred provider's identity being unchanged, and resets otherwise.

**5. Alias-only labels are ambiguous and under-specified** — Agreed on both counts. A composite alias can carry several named pools whose independent controllers would write the same series and aggregate unrelated observations. Labels are now `model`, `pool`, and a bounded `role` of `preferred` or `alternate` — which is also what makes per-provider comparison possible without unbounded provider identities, given the recorder deliberately runs without eviction. The share gauge carries no `role`, since one controller governs one pool.

Still design-only. Implementation of step 2 (observation: the uncensored histogram and breach counter, labelled by role, with `[DONE]` excluded from counting as a first token) is next on this branch, at which point the PR title will move to `feat`.

PR #1828, opened 2026-09-21
fix(underway): ordered claim index, bounded task retention, purge script

pjb157:
## Why

Underway 0.2.0's task claim orders live tasks by `(priority DESC, created_at, id)` and takes one row, but no index matches that order. Every poll therefore collects the queue's live rows and sorts them, on every worker, so a backlog makes the claim itself the bottleneck and the database pays for it continuously.

Separately, nothing ever removes finished tasks. The crate honours each task's `ttl` only when its deletion routine is running, and that routine is a single unbounded `DELETE`, so the task table grows for the life of an installation.

## What

**Indexes** (`dwctl/underway-migrations`, add/reindex/validate/cleanup sequence per `docs/migrations.md`):
- `idx_task_claim` on `(task_queue_name, priority DESC, created_at, id) WHERE state IN ('pending','in_progress')`. A custom plan proves the partial predicate and walks the index in claim order, stopping at the first eligible task. A generic plan cannot prove it (the states are bind parameters) and keeps using `idx_task_queue_state`, which stays for that reason and as the completion-side lookup. The existing plan test now asserts the expected index per plan-cache mode.
- `idx_task_created_at`, so retention deletes are a bounded range scan.
- Drops a hand-built pending-only `(task_queue_name, created_at)` index that some installations carry. No release created it, the claim cannot use it, and `idx_task_claim` covers its lookup.

**Task retention daemon** (`background_services.task_retention`, on by default): deletes expired tasks oldest-first in batches, each its own short transaction, with a pause between batches. Expired means not `in_progress`, older than `min_age_days` (default 14, the crate's default `ttl`), and past the row's own `ttl`. An advisory lock collapses concurrent sweeps to one; `SKIP LOCKED` keeps it off rows a worker is claiming. `task_attempt` rows follow through the schema's `ON DELETE CASCADE`. Documented in `docs/src/reference/configuration.md`.

**`scripts/purge_underway_tasks.sh`**: the same batched delete from the shell, for draining a large existing backlog in the background before the daemon takes over. Configurable batch size, pause, age floor and queue filter; per-batch lock and statement timeouts; `DRY_RUN=1` reports what qualifies; refuses to run without `idx_task_created_at`.

## Rollout notes

- Index builds are `CONCURRENTLY`; the usual Underway extension prerequisites in `docs/migrations.md` apply.
- The previous release's workers use the new claim index too, so this is additive.
- Deleted rows only return space to the OS after `VACUUM (FULL)` or `pg_repack`; a normal autovacuum makes the pages reusable.

## Tests

- `cargo test -p dwctl --lib -- migrations:: task_retention::` (26 passed): plan assertions for both plan-cache modes, retention semantics (state, floor, per-row ttl, batch bound, oldest-first, cascade, lock yield), and the retention index plan.

<a href="https://cubic.dev/pr/doublewordai/control-layer/pull/1828?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


pjb157:
Addressed the review in the follow-up commit:

- **Advisory lock lapsing between batches** (cubic, `task_retention.rs:86`): the sweep now takes a session-level `pg_try_advisory_lock` on a detached connection held for the whole sweep and closed on every exit path; each delete batch still commits on its own. New test `sweep_holds_its_lock_across_batches` competes during a pause and is refused; the existing lock test also checks the lock is released afterwards.
- **`purge_expired_batch` transaction claim** (cubic, `task_retention.rs:33`): it now takes the pool and owns its transaction.
- **`min_age_days` overflow** (cubic, `config.rs`): bounded to 36,500 days in `Config::validate` (with `batch_size >= 1`), and the conversion clamps as well.
- **Drop of an index owned by another role** (cubic, migration `…090`): a transactional guard migration `…085` now precedes it and fails with the exact `ALTER INDEX … OWNER TO` to run when the migration role cannot drop the index, instead of an opaque ownership error mid-target.
- **Validation coverage for the new indexes** (Copilot, cubic, migration `…030`): `underway_extensions_repair_interrupted_indexes` now also repairs invalid `idx_task_claim` / `idx_task_created_at`; new tests reject wrong key order, missing `DESC`, a pending-only predicate, no predicate, and a wrong retention index.
- **Queue names pasted into SQL** (Copilot, cubic, script): the filter is now a psql variable interpolated as `:'queues'` via stdin (psql does not interpolate in `-c`), verified with a queue named `o'reilly`.
- **Negative / malformed numeric inputs** (cubic, script): all numeric env vars are validated before use.
- **Name-only index check** (Copilot, cubic, script): now checks `indisvalid`, `indisready`, btree, and the exact `(created_at)` definition, same as the validation migration.
- **Dry run without timeouts** (cubic, script): the dry run goes through the same guarded path as the batches.

